PR #191, opened 2014-01-05
Update Maths.wurst

edo494:
Optimized functions min(int, int), min(real, real), max(int, int) and max(real, real)

Now, it just performs return instead of assign and return(don't know if that would get optimized, most likely yes but still)


Cokemonkey11:
@edo494 you have merge conflicts

Frotty:
@peq Will this ever happen ? :D

peq:
It might happen, but maybe we should close this pull-request for now.

PR #346, opened 2015-02-28
rewrote TimedLoop, add returns to add|removeCondition in Trigger, Assert (wasn't able to use Wurstunit that's why i wrote it), fixed/tested OnUnitEnterLeave and UnitIndexer

Ruk33:


Ruk33:
fixed initialization on OnUnitEnterLeave package, also DamageDetection rewrited mainly because to avoid the use of one giant trigger

Cokemonkey11:
You have merge conflicts, you should also squash your commits and use a separate branch

Ruk33:
uhmm, i think the std library should be separated from the compiler...

Frotty:
it will be, when we have some sort of dependency management.

peq:
> Also please create smaller pull requests, which can be reviewed easier.

This would be good. Create one branch per feature/ pull request to do this.

PR #391, opened 2015-09-03
Rewrote from scratch.

muzzel:


peq:
This does not look like it is backwards compatible with the old library, so maybe it is better to publish this as a separate library? Or a different package?

I have to work on integrating some dependency management into Wurst.

Frotty:
Who used the old crappy library? :palm_tree:

Crigges:
I don't

peq:
It is not just about this package: In general we have to make the standard library smaller and move more things (or everything) to external, versioned dependencies. I don't like the current situation where adding a function to the standard library can break someones code just by upgrading Wurst. 

But yeah, as long as I did not implement the tools to make this work we can also merge this into the standard library.

Frotty:
So this won't be a problem for the next 1-2 years :shell:

peq:
let's see :dolphin:

muzzel:
I thought about providing an additional constructor with the same signature as the old one, but in the oly library there are unused constructor arguments and I definitely wont add sh*t like that only for compatibility.

PR #401, opened 2015-09-09
Added operatoroverloading for dot product

Crigges:
Not sure if you think this is right
blabalbalbalbal

### DIFF
--- a/Wurstpack/wurstscript/lib/math/Vectors.wurst
+++ b/Wurstpack/wurstscript/lib/math/Vectors.wurst
@@ -39,8 +39,12 @@ public function vec3.op_mult(real factor) returns vec3
 /** * Operator */	
 public function real.op_mult(vec3 v) returns vec3
 	return vec3(v.x*this, v.y*this, v.z*this) 
+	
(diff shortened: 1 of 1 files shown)


peq:
I don't think it is a good idea to use the multiplication sign for the dot-product. Multiplication usually has the type `a -> a -> a`. Also it could be confused with the cross-product or with pairwise multiplication.

We already have the `dot` extension method for computing the dot product, which I find easier to read. Operator overloading should only be used if the meaning is known and used by everyone in the domain. For the dot product there exist other notations like `<a,b>`.

Other languages with operator overloading also mainly use named functions instead of operator overloading, e.g. [Haskell uses a special function ](https://hackage.haskell.org/package/AC-Vector-1.2.2/docs/src/Data-Vector.html#vdot), the Scala library [Breeze as well ](http://www.scalanlp.org/api/breeze/#breeze.linalg.Vector), the C++ library [Boost](http://www.boost.org/doc/libs/1_42_0/libs/numeric/ublas/doc/operations_overview.htm#23innerouterandotherproducts).

Issue #355, opened 2015-04-22
After importing sounds Wurst stops working


ENAleksey:
I imported sounds in my map, now error occurred when I trying to save it:
![screenshot_1](https://cloud.githubusercontent.com/assets/12061580/7267397/edacf00e-e8d3-11e4-939a-f69ecb7eeecc.jpg)

peq:
Can you please upload the sound somewhere? Is it a wav or mp3 sound?

@Crigges : do you have time to look at this bug?

ENAleksey:
It is a wav. file. It happens with any sounds, I tried to save map without Wurst and it worked.

ENAleksey:
With .mp3 files it works without error.

peq:
@ENAleksey  : Can you please upload the wav file (or even better: a complete map with a wav file) which does not work for you? We are having problems reproducing the error. Thank you.

ENAleksey:
[Test map](https://mega.co.nz/#!ec1zXYRA!iqBTnPlN7v3WaljEMZMdm0d8CfLqFhAmv9bnHX5hXsA)

ENAleksey:
After importing [this texture](https://mega.co.nz/#!7Bdk2A4I!PuMLbopQM-qBbqSVSeQKUzGyMWCvh3j6ArxRJlHiQx8), error occurred too:
![screenshot_1](https://cloud.githubusercontent.com/assets/12061580/7471951/88255cda-f346-11e4-93cd-38136be5af09.jpg)

peq:
@Crigges is working on a new version of the MPQ library. Maybe this problem would be fixed by it as well.

Until this is done, you could try to import stuff with the import tool. (Just put all imports into a folder named `imports` as done in the HotN (https://github.com/Crigges/HotN/tree/master/HotN) project.

If this does not work either, I could temporarily add optional support for external mpq editors again, but this had a lot of other problems.

ENAleksey:
I tried to use import tool, but without success. I need these files, so what should I do?

Frotty:
Not use .wav files. Convert them to .mp3 - if you want 3d sound, use https://github.com/peq/WurstScript/blob/master/Wurstpack/wurstscript/lib/small%20helpers/Simulate3dSound.wurst

ENAleksey:
What about textures? I tried to remove the alpha channel, use compressed textures, but it is not working. Helps only reducing size of the textures, but I need the original size.

peq:
[...]

Issue #327, opened 2015-01-24
Report: JMpq fails if file from listfile is not found


peq:
```
File Not Found 
de.peeeq.jmpq.HashTable.getBlockIndexOfFile(HashTable.java:84)
de.peeeq.jmpq.JMpqEditor.<init>(JMpqEditor.java:79)
de.peeeq.wurstio.mpq.Jmpq2BasedEditor.<init>(Jmpq2BasedEditor.java:28)
de.peeeq.wurstio.mpq.MpqEditorFactory.getEditor(MpqEditorFactory.java:10)
de.peeeq.wurstio.Main.main(Main.java:122)
```

peq:
hope that fixes the problem, but need some feedback from reporter

Frotty:
What about the other threleven issues pöq.?

peq:
Most of them are just minor issues. There is nothing with high priority label and everything with normal priority is also marked as only an enhancement. There is only one other reproducible issue marked as "bug". 

You should just join IRC again and tell me what is important :D

peq:
Fixed by JMpq-v3

Issue #323, opened 2014-12-14
Report: JMpqException: Unsupported compression algorithm: -127


peq:
See http://peeeq.de/wursterrors.php?id=213

peq:
Fixed by JMpq-v3

Issue #321, opened 2014-11-28
Report: JMpqException in Compilation-Server: Could not overwrite the orginal mpq: null


peq:
```
de.peeeq.jmpq.JMpqException: Could not overwrite the orginal mpq: null

Version: 1.5.0.0-jenkins-Wurst-783

de.peeeq.wurstio.mpq.Jmpq2BasedEditor.close(Jmpq2BasedEditor.java:54)
de.peeeq.wurstio.Main.main(Main.java:130)
de.peeeq.wurstio.compilationserver.WurstServer.lambda$0(WurstServer.java:89)
de.peeeq.wurstio.compilationserver.WurstServer$$Lambda$11/1848402763.run(Unknown Source)
java.awt.event.InvocationEvent.dispatch(Unknown Source)
java.awt.EventQueue.dispatchEventImpl(Unknown Source)
java.awt.EventQueue.access$400(Unknown Source)
java.awt.EventQueue$3.run(Unknown Source)
java.awt.EventQueue$3.run(Unknown Source)
java.security.AccessController.doPrivileged(Native Method)
java.security.ProtectionDomain$1.doIntersectionPrivilege(Unknown Source)
java.awt.EventQueue.dispatchEvent(Unknown Source)
java.awt.EventDispatchThread.pumpOneEventForFilters(Unknown Source)
java.awt.EventDispatchThread.pumpEventsForFilter(Unknown Source)
java.awt.EventDispatchThread.pumpEventsForHierarchy(Unknown Source)
java.awt.EventDispatchThread.pumpEvents(Unknown Source)
java.awt.EventDispatchThread.pumpEvents(Unknown Source)
java.awt.EventDispatchThread.run(Unknown Source)
```

http://peeeq.de/wursterrors.php?id=211

Please post here if you have this problem, I think I need some more information to reproduce this bug.

peq:
Fixed by JMpq-v3

PR #408, opened 2015-09-24
Added JMpq-v3 support

Crigges:
Important features:
-Decreased ram usage significantly
-Can open and rebuild mpqs including files that are compressed using other compression algorithms as zlib
-Much more stable implementation and cleaner code -> easy to add features
-Better resulting filesize

I've been testing Jmpq-v3 the last few days on different maps without problems so far. 
Since v3 was intended to replace v2 iam not going to fix any bug in v2. So the tickets: #327 #321 #355 and #323 can be closed.


peq:
I want to test it with a few maps myself, before merging this. Probably I can do it this evening.

peq:
Tested with 2 maps and it worked. Thus, by induction it works for all maps.

PR #426, opened 2016-01-30
(Needs discussion and validation) boolean.xor(boolean)

Cokemonkey11:
I'm not entirely sure wurstscript doesn't already have xor, but I didn't see it anywhere.

I'm also not so sure this is the best implementation. Maybe someone else can convince me that `boolean.xor(boolean)` looks good, or otherwise?


Cokemonkey11:
[10:45] <lep_> a != b

herpderp

PR #432, opened 2016-02-19
Fixed leave-detection

muzzel:
- fixed leave detection in general (leaving units were not detected due to wrong event response)
- fixed leave detection on preplaced units (wasnt working at all, wrong event response)
- removed the minimal unit indexer, it doesnt belong here and there is a real unit indexer in the stdlib which is 100% compatible


Frotty:
Thanks :heart:

Issue #305, opened 2014-08-11
Remove old dummy recycler


peq:
And rename `NewDummyRecycler`, if there is no old one it does not make sense to only have a new one ;)

Frotty:
see https://github.com/peq/WurstScript/pull/435

PR #435, opened 2016-03-09
Recycler patch for #305

Frotty:
NewDummyRecycler -> DummyRecycler


Frotty:
@peq pls merge if appropriate

peq:
Another breaking change... But I guess that this package is barely used as a direct import in maps, so it's probably ok.

PR #436, opened 2016-03-18
[WIP] Gradle integration

Frotty:


Frotty:
Anything new regarding this topic @peq ?

peq:
I tried to get it to work with the eclipse-plugin (using Wuff), but I failed to do so.

To switch to Gradle the following things have to work as they do now:

- compile the eclipse plugin to create the update-site from the commandline
- run/debug the eclipse plugin from eclipse
- easily update the version number in a single place

Moreover I could not get it to compile with the the Eclipse Java compiler (EJC). The normal JDK compiler is really slow on Wurst, because there is a performance bug in Java 8 in overloading resolution and Wurst uses a lot of overloading internally.

If you manage to fix these things, I would accept the pull request, but there are more important things to do for Wurst.

Issue #421, opened 2015-11-04
Launching map gets stuck at 24% "Launching delegate..." and Eclipse uses 650+ MB RAM


Cokemonkey11:
![image](https://cloud.githubusercontent.com/assets/1418966/10954852/c9b5f786-8347-11e5-851e-c3547e771817.png)

It happens to be a new map that has not yet been successfully run by wurst.

I launch it by right clicking the w3x file in eclipse and selecting "run as Wurst".

Crigges:
My eclipse uses 2GB Ram if it is hungry so be be happy with 700MB

The launch seems to be stucked in a infinity loop. Better show us the console output instead of the progressbar.

Cokemonkey11:
I cleaned and rebuilt wurst, and deleted the `wurst/war3map.j` file and rebuilt by saving the map in wurstpack, then tried doing again "run as > wurst":

![image](https://cloud.githubusercontent.com/assets/1418966/10983071/7f43fef2-8408-11e5-91e4-e7f564197936.png)

Wurst console only shows:

```
building project rise-of-daolmir, please wait ...
```

Crigges:
Looks like you are using a old wurstscript version.

Cokemonkey11:
@Crigges nope, it's the newest clean build from peq/Wurstscript/master

Crigges:
Start eclipse via the eclipsec.exe and show us the cmd output.

Cokemonkey11:
When I right click map and select "run as > wurst" I get different errors sometimes, it seems. In this particular example:

![image](https://cloud.githubusercontent.com/assets/1418966/11020893/215be3b0-8627-11e5-9f79-7dc6ac50c002.png)

And here is the output of eclipsec.exe -

https://gist.github.com/Cokemonkey11/8b9494bd99540559bc4d

Crigges:
JMpq-v2 is used instead of v3. This was fixed a few commits ago. Check the stacktrace on your own:
The stacktrace points to:
  at de.peeeq.eclipsewurstplugin.console.WurstREPL.compileScript(WurstREPL.java:641)

This is the line:
https://github.com/peq/WurstScript/blob/master/EclipseWurstPlugin/src/de/peeeq/eclipsewurstplugin/console/WurstREPL.java#L641

So since line 641 is ```}``` your eclipse plugin version is outdated.

Cokemonkey11:
Crigges, thanks a lot for the help so far.

[...]

PR #458, opened 2016-07-16
Updated jmpq3

Crigges:
The old old jar was causing some bugs, which should be fixed now. See #421


Frotty:
"should be" or did you actually test it for once?

Crigges:
I was able to reproduce the bug inside the jmpq libary. After the fix the same piece of code was working. I aswell tested the new .jar inside the we without any problems.

Frotty:
This breaks maps for ghost bots again, pls fix.

Crigges:
Ghost can host maps compiled with the new jar now.

Frotty:
I will confirm this and then merge.

PR #459, opened 2016-07-16
Fix ClosureForGroups.wurst

### LINKED ISSUE #456: How to deal with ForGroupCallback.callback on invalid object Error
For example:

forgroup-bug

And here is every instance of ClosureForGroups used in my map:

		guard = null // Global variable.
		forUnitsInRange(vec2(u.getX(), u.getY()), RADIUS, (unit iter) -> begin
			if iter.isAlive() and isGuardDevil(iter) and iter.hasAbility(ID_ABILITY_UNBREAKABLE)
				guard = iter
			end
		end)

and

			hasUnit = false // Global variable.
			forUnitsInRange(vec2(leviathans[pId].getX(), leviathans[pId].getY()), RADIUS, (unit iter) -> begin
				if iter.isAlive() and IsUnitEnemy(iter, owner)
					hasUnit = true
				end
			end)

and

			forUnitsInRange(vec2(tU.getX(), tU.getY()), RADIUS, (unit iter) -> begin
				if iter.isAlive() and IsUnitEnemy(iter, p)
					iter..issueTargetOrder("attack", tU)
					    ..addAbility(DAMAGE_REDUCTION_ID)

					addTaunt(iter, tU, DURATION)
				end
			end)

and

		willSacrifice = false // Global variable.
		forUnitsInRange(vec2(u.getX(), u.getY()), RADIUS, (unit iter) -> begin
			if iter.isAlive() and iter.getTypeId() == ID_THALL_BOX and iter.getOwner() == u.getOwner() and iter.hasAbility(ID_ABILITY_SACRIFICE)
				willSacrifice = true
			end
		end)

and

		splendorCheck = false // Global variable.
		forUnitsInRange(vec2(u.getX(), u.getY()), RADIUS, (unit iter) -> begin
			if iter.isAlive() and iter.getTypeId() == ID_THALL_BOX and iter.getOwner() == u.getOwner() and iter.hasAbility(ID_ABILITY_SPLENDOR)
				splendorCheck = true
			end
		end)

		return splendorCheck
	end

Frotty:
I think this should fix #456

### DIFF
--- a/Wurstpack/wurstscript/lib/small helpers/ClosureForGroups.wurst
+++ b/Wurstpack/wurstscript/lib/small helpers/ClosureForGroups.wurst
@@ -16,7 +16,7 @@ function pushCallback(ForGroupCallback c)
 	tempCallbacks[tempCallbacksCount] = c
 	tempCallbacksCount++
 
-function popCallback() returns ForGroupCallback
+function popCallback()
 	tempCallbacksCount--
 	destroy tempCallbacks[tempCallbacksCount]
 
--- a/Wurstpack/wurstscript/lib/small helpers/ClosureForGroups.wurst
+++ b/Wurstpack/wurstscript/lib/small helpers/ClosureForGroups.wurst
@@ -18,7 +18,7 @@ function pushCallback(ForGroupCallback c)
 
 function popCallback() returns ForGroupCallback
 	tempCallbacksCount--
-	return tempCallbacks[tempCallbacksCount]
+	destroy tempCallbacks[tempCallbacksCount]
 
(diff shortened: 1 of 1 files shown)


peq:
This looks like it does not compile. I think it is better to have the fixed library in a new package, as this changes the behavior of the library. 

If you use it with closures, you would expect the closure to be destroyed, but if you pass a normal object, destroying it might be a surprise.

Frotty:
it does not change it ? The return value of pop is never used

peq:
The name of the library suggests that this should be used with Closures. As this use case is not affected, I think it is ok to change the behavior of the package in this case.

PR #469, opened 2016-07-25
[stl] add texttag critical strike macro

Cokemonkey11:
Credits: http://www.wc3c.net/showthread.php?t=101963

### DIFF
--- a/Wurstpack/wurstscript/lib/handles/Texttag.wurst
+++ b/Wurstpack/wurstscript/lib/handles/Texttag.wurst
@@ -42,6 +42,17 @@ public function texttag.setVisibility( boolean flag )
 public function texttag.destr()
 	DestroyTextTag(this)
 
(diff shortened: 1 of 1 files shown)


peq:
I think it is strange that this should be an extension function on `texttag`. It is a very specific use case and should probably be in a separate package like `StandardTexttags`.

Frotty:
Indeed. Together with others like bounty, exp, miss, etc.

Cokemonkey11:
I can make it right now since you both seem to be in agreement?

For the API, how do you feel about

```
texttagCriticalStrike
texttagShadowStrike
texttagMiss
```

etc

peq:
I like it when function names have a verb, like `createCriticalStrikeText(...)`.

Cokemonkey11:
standby for new pr

PR #470, opened 2016-07-25
[stl] add standard text tags package

Cokemonkey11:
This suffers from a bit of code duplication but it is not too bad.

One potential improvement would be to extract the colors into
more reuseable vec3s in Colors package.

(not tested)

### DIFF
--- a/Wurstpack/wurstscript/lib/small helpers/StandardTextTags.wurst
+++ b/Wurstpack/wurstscript/lib/small helpers/StandardTextTags.wurst
@@ -1,64 +1,51 @@
 package StandardTextTags
-    constant real FONT_SIZE  =  0.024 / 0.0023
-    constant real SIGN_SHIFT = 16.0
+    constant FONT_SIZE  =  0.024 / 0.0023
+    constant OFFSET     = vec2(16., 0.)
+
+
+    function _standardTextTag(vec2 pos, string text) returns texttag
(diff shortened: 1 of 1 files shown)


Cokemonkey11:
@Frotty do you plan to review this?

Wurstpack/wurstscript/lib/small helpers/StandardTextTags.wurst
```diff
@@ -0,0 +1,76 @@
+package StandardTextTags
+    constant real FONT_SIZE  =  0.024 / 0.0023
```
Frotty: Value based on?

Frotty:
There could probably be some preset function for texttags itself that these functions call so it's not the same extension funcs in every function.
Please add a one-line of hotdoc to each func then I will merge

Wurstpack/wurstscript/lib/small helpers/StandardTextTags.wurst
Cokemonkey11: http://www.wc3c.net/showthread.php?t=101963

Cokemonkey11:
@Frotty okay, this should be good. Let me know if it's ready and I will squash.

peq:
I put the StdLib into a separate repo, please create the same pull request over there:  https://github.com/peq/wurstStdlib

PR #473, opened 2016-07-28
[stl] fix error in linked list iterator destroy method

Cokemonkey11:


Wurstpack/wurstscript/lib/data structures/LinkedList.wurst
```diff
@@ -151,8 +151,8 @@ public class LLIterator<Q>
 		if current != dummy
 			current.next.prev = current.prev
 			current.prev.next = current.next
-			current = current.prev
 			destroy current
+			current = current.prev
```
peq: you should not use an object after it is destroyed. This might give compilation errors in future versions.

Frotty: better than a bug ;) i fixed it in stdlib2 the same way some time ago.
I hope
````
			let temp = current
			current = current.prev
			destroy temp
````
gets inlined then

Issue #349, opened 2015-04-16
No documentation for using vjass maps with wurst compiler


blakehawkins:
I recall seeing an explanation somewhere of how to use wurst with an existing vjass project - extracting all the vjass triggers into wurst packages and compiling via wurst, but the manuals don't contain info on this.

Such a nice feature should be documented.

Cokemonkey11:
whoops, I keep accidentally posting here with my other alias - this is me

peq:
This is an experimental feature together with Jurst, but I agree that there should be some documentation.

Cokemonkey11:
It seems to work quite well. One suggestion would be to write something in wurst console when beginning/ending the process successfully. I thought the feature was broken until I refreshed my project.

Cokemonkey11:
@peq I'm happy to look at this week. I'm thinking that a section could be added to gh-docs branch which includes a short explanation of:

* how wurst behaves when compiling a map with existing vjass triggers
* how the 'extract to jurst' feature works and some steps to use it

Let me know if this sounds good and I'll get on it.

peq:
Sounds good. Although the extract to Jurst feature is not yet included in vscode. Just don't put too much effort in explaining eclipse specific things.

Frotty:
I think this is properly addressed now (tutorials on side and on hive)

PR #493, opened 2016-08-22
add some documentation about jurst and managing legacy maps

Cokemonkey11:
Related #349 

I think there's some language which is not quite right ("translate vjass"?) Let me know what you think.


peq:
Please try to keep one sentence per line. It is easier to read diffs that way.

> the Wurstpack World Editor will parse jass/vjass, in addition to compiling your Wurst code.

It will only parse jass unless you enable the vjass compiler.

> translating existing vJass to Jurst automatically

There is no automatic translation. The command just extracts the trigger code as it is and puts it into files.

Cokemonkey11:
I'll not have time anymore to make the required changes. @Frotty can you take this over please?

Cokemonkey11:
@peq I've fixed this up.

Frotty:
ty

PR #573, opened 2017-11-14
Changed order of compiletime execution (WIP)

peq:
Old rule:

- first all compiletime expressions are executed
- then compiletime functions
- no other order contraints specified

New rule:
- compiletime expressions and functions interleaved
- guaranteed that compiletime stuff in imported packages is executed first
- inside a package it should be top down (according to AST order)

Still untested, so do not merge before testing it.

### DIFF
--- a/de.peeeq.wurstscript/parserspec/jass_im.parseq
+++ b/de.peeeq.wurstscript/parserspec/jass_im.parseq
@@ -84,7 +84,7 @@ ImFlatExpr =
 	| ImClassRelatedExpr
 	| ImConst
 	| ImGetStackTrace()
-	| ImCompiletimeExpr(de.peeeq.wurstscript.ast.Element trace, ImExpr expr)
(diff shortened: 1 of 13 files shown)


Frotty:
Okay, I added a test in Colors.wurst that simply accesses an array init in init block.
```
@test
public function test()
	print("hex5: " + hexs[5])
```
Still same error:
```
FAILED with exception:
	You encountered a bug in the interpreter: java.lang.ClassCastException: de.peeeq.wurstscript.intermediatelang.ILconstNull cannot be cast to de.peeeq.wurstscript.intermediatelang.ILconstAddable
```

peq:
That is expected. Init blocks are still not executed at compiletime, because they usually use natives which are only available in the game.

You have to change it to something like this if you need the init at compiletime:

```
init
    doInit()

@compiletime
function doInit() 
    ...
```

PR #583, opened 2017-11-21
Win files fix

Frotty:
This fixes https://github.com/wurstscript/WurstScript/issues/582 for me, however I'm not sure if it is elegant. JumpToDecl now keeps working and the errors don't appear.
I debugged and found out that the problem was that ModelManagerImpl caches e.g. common.j in build folder with empty content, i.e. hash 0. I think this is due to ModelManager & BufferManager having independent caching, but I'm not sure.
Also there was some "didOpen" "didClose" right after each other in the log, indicating some file being opened and closed, idk why, but it's gone now as well.

In general the whole file/model handling should probably be refactored at some point.
Things like WurstWorkspaceService which work solely on BufferManager and not ModelManager seem to not even be used?

This still isn't perfect of course. Another thing I noticed now if jumping into common.j, then back, modifying and saving the original file, many warnings are being swallowed and only return after reload.

### DIFF
--- a/de.peeeq.wurstscript/src/main/java/de/peeeq/wurstio/languageserver/BufferManager.java
+++ b/de.peeeq.wurstscript/src/main/java/de/peeeq/wurstio/languageserver/BufferManager.java
@@ -21,7 +21,7 @@ public class BufferManager {
         return getBuffer(uri);
     }
 
-    private String getBuffer(WFile uri) {
+    public String getBuffer(WFile uri) {
         StringBuilder sb = currentBuffer.get(uri);
         if (sb == null) {
             return "";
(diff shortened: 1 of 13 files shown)


Frotty:
tests pass locally for me, dunno what travis needs.
HT and DD have tested this and confirmed it fixes the issue for them.

IgorSamurovic:
Can confirm, I am no longer getting this issue, and if I do I will report it.

peq:
I took the main finding from this pull request (buffer could be null) and fixed it directly in the buffer manager  abee3ef6ce315a39f27cccd2ece6721176addbf4

PR #608, opened 2017-12-23
fix the slashes even more.

Trokkin:
The fix that Frotty has made is broken, so that files in directories has zero size somehow, still the entire map seems to contain the data, but unavailable to WC3.
I haven't tested the changes, but it should work.


Frotty:
thanks, but instead of 2 calls I will just change the local.

PR #620, opened 2018-02-03
Pretty printer alpha

karlek:
This is the alpha implementation of the pretty printer; I want to say thank you to @peq who wrote the initial structure of the printer, and to @Frotty who has answered my many specific questions about the wurst parser.

---

The current state of the pretty printer is that it correctly reconstructs wurst files. However, it doesn't retain neither explicit newlines inside statements nor line comments. Line comments aren't saved during parsing, which needs to be implemented first. Retaining explicit newlines is a little more difficult to implement, one thought is by using `WPos` of the element and compare it to the `WPos` of its parent and thereby inserting explicit newlines. 

---

Future features:
* Variable alignment
* Precedence aware ExprBinary parsing, i.e: insert whitespace to highlight algebraic precedence. Examples: (1*5 + 3) and (3 + 1)
* Newline preserving prettying.
* Needs more tests

---

_Some more questions:_

Why is WurstDoc a modifier? It feels wrong compared to other modifiers such as `public`, `constant`, etc.

How does multiple module usages look like?
```go
// Like this?
use A, B
// Or like this:
use A
use B
```

I wrote a script that got the AST for each file in the WurstStdlib2 and didn't encounter any of the following tokens:
* `ModuleInstanciation`
* `WPackages`, can you have more than one `package` in a file?
* `WBlock` is it something like this:
```go
// Statements.
{
   // Statments.
}
```
* `VisibilityPublicread`
* `VisibilityDefault`


de.peeeq.wurstscript/src/main/java/de/peeeq/wurstscript/WurstParser.java
```diff
@@ -194,7 +194,7 @@ public CompilationUnit emptyCompilationUnit() {
     }
 
     private void removeSyntacticSugar(CompilationUnit root, boolean hasCommonJ) {
-        new SyntacticSugar().removeSyntacticSugar(root, hasCommonJ);
+        // new SyntacticSugar().removeSyntacticSugar(root, hasCommonJ);
```
karlek: Shouldn't exist in the final version 🌵

de.peeeq.wurstscript/src/main/java/de/peeeq/wurstscript/attributes/prettyPrint/Main.java
```diff
@@ -0,0 +1,139 @@
...
+
+import static de.peeeq.wurstscript.utils.Utils.printElement;
+
+
+public class Main {
+
+    /**
+     * @param args
+     */
+    public static void main(String[] args) {
```
Frotty: shouldn't have it's own main func.
Instead have runarg for external use and WorkerTask for vscode integration

de.peeeq.wurstscript/src/main/java/de/peeeq/wurstscript/attributes/prettyPrint/Main.java
```diff
@@ -0,0 +1,139 @@
...
+        lastElement = e;
+        check(e, indent);
+        lastElement = null;
+        for (int i = 0; i < e.size(); i++) {
+            walkTree(e.get(i), indent+1);
+        }
+    }
+
+
+    private static void prettyWalk(String rootPath) {
```
Frotty: should be replaced with something like:
```
        Files.walk(Paths.get(rootPath))
                .filter(p -> p.toString().endsWith(".wurst"))
                .forEach(p -> prettyWalk(p))
```

de.peeeq.wurstscript/src/main/java/de/peeeq/wurstscript/attributes/prettyPrint/Main.java
```diff
@@ -0,0 +1,139 @@
...
+                if (extension(path).equals("wurst")) {
+                    pretty(path);
+                }
+            }
+        }
+    }
+
+    // Note: only works with single extension files, i.e it doesn't work with:
+    // asdf.tar.gz
+    private static String extension(String filename) {
```
Frotty: should be obsolete when using above FIle.walk

de.peeeq.wurstscript/src/main/java/de/peeeq/wurstscript/attributes/prettyPrint/Main.java
```diff
@@ -0,0 +1,139 @@
...
+import java.io.File;
+import java.io.FileReader;
+import java.io.BufferedReader;
+import java.io.FileInputStream;
+import org.apache.commons.lang.StringUtils;
+
+import static de.peeeq.wurstscript.utils.Utils.printElement;
+
+
+public class Main {
```
Frotty: Should probably be renamed and provide public access to these functions, so they can be used from cli and from vscode

de.peeeq.wurstscript/src/main/java/de/peeeq/wurstscript/attributes/prettyPrint/Main.java
```diff
@@ -0,0 +1,139 @@
...
+        Spacer spacer = new MaxOneSpacer();
+        StringBuilder sb = new StringBuilder();
+        cu.prettyPrint(spacer, sb, 0);
+
+        System.out.println(sb.toString());
+    }
+
+    private static String readFile(String filename) {
+        String everything = "";
+        try {
```
Frotty: use try with resources

de.peeeq.wurstscript/src/main/java/de/peeeq/wurstscript/attributes/prettyPrint/PrettyPrinter.java
```diff
@@ -3,10 +3,82 @@
 import de.peeeq.wurstscript.ast.*;
 import org.apache.commons.lang.StringUtils;
 
+import static de.peeeq.wurstscript.utils.Utils.escapeString;
+import static java.lang.Math.min;
+
 public class PrettyPrinter {
 
+    private static void commaSeparatedList(Element e, Spacer spacer, StringBuilder sb, int indent) {
```
Frotty: consider replacing with Utils.join

de.peeeq.wurstscript/src/test/java/tests/prettyprint/PrettyPrintTest.java
```diff
@@ -1,40 +1,119 @@
...
+        return sb.toString();
+    }
+
+    private String expectedFile(String filename) throws IOException {
+        StringBuilder sb = new StringBuilder();
+        File inFile = new File(filename);
+        return Files.toString(inFile, Charsets.UTF_8);
+    }
+
+    private void visitEvery(String rootPath) throws IOException {
```
Frotty: same as in Main

peq:
Hi,

thanks for your work!

I added a suggestion for tracking comments on this branch: https://github.com/wurstscript/WurstScript/tree/pretty-printer-comments

Maybe you can work with this as a starting point? 

The approach there is to store all comments in a list and then add them in the printer based on their position (line/column) in the source code. For example, if we print an element with position 100, the printer will first print all comments before position 100.

One problem with this is, that the printer cannot change the order of elements, since that would mess up the comment order. To support reordering, some smarter mapping from comments to AST elements could be done in a preprocessing step.

Another problem is, that comment positions are not kept precisely, e.g. it is not possible to distinguish `1 /* comment */ + 2` and `1 + /* comment */ 2`. 
If we want this, we probably need a different approach. Maybe we can have a look at how http://scalameta.org/ does it, since they apparently support this. However, I am a bit afraid that a more detailed representation of the source code would lead to a higher memory usage and some Wurst users already have reported performance problems related to that.

karlek:
Hi,

It has been really fun working on this and I've grown more accustomed to wurst as a pleasant side effect.

Well, there's a few parts in the pretty printer that will often change the token order, the best example is `ClassDef`. Since `ClassDef` exposes `.getConstructors()`, `.getInnerClasses()`, etc. as methods we lose the ordering of the their tokens.

I tried your branch, but there seems to be some problems with the parser. 

2018-03-06-141445_2560x1440_scrot

peq:
> I tried your branch, but there seems to be some problems with the parser.

Oh, I somehow missed a `continue`. Should be fixed now on the branch.

However, after thinking a bit more about it, I guess we should use the approach to store the tokens attached to the AST nodes. That is more precise and does not have the reordering problem. To keep memory usage low, we can make the parser configurable and only generate the extra information if we need it for detailed pretty printing. I think I'm gonna try that variant on the weekend.

Frotty:
Okay so what is our plan with this?
We want to keep comments, so in it's current state it is somewhat unacceptable.
@peq @karlek

karlek:
@Frotty, currently WurstScript lacks complete parsing of comments (line comments, and block comments). 

As @peq pointed out:

> However, after thinking a bit more about it, I guess we should use the approach to store the tokens attached to the AST nodes. That is more precise and does not have the reordering problem. To keep memory usage low, we can make the parser configurable and only generate the extra information if we need it for detailed pretty printing...

However, if @peq doesn't have the time, maybe I can take a stab at it. However, I don't have any experience with ANTLR so I'll need to ask a lot of questions if I do.

peq:
I would suggest the following: 
1. I adapt the parser, so that it (optionally) stores comments and maybe other tokens as well.
2. @karlek you adapt the pretty printer to include the comments

peq:
I started adding the comments to the AST on the `karlek-pretty-printer` branch: https://github.com/wurstscript/WurstScript/tree/karlek-pretty-printer



To get the comments, you can use the functions `getCommentsBefore` and `getCommentsAfter` as I did here: https://github.com/wurstscript/WurstScript/blob/karlek-pretty-printer/de.peeeq.wurstscript/src/main/java/de/peeeq/wurstscript/attributes/prettyPrint/PrettyPrinter.java#L68-L100

See `testComments` in `PrettyPrintTest` for a testcase that is now partially working. The test also prints the AST to see where the comment tokens are stored.

Frotty:
Any chance of you looking into this again in the near Future @karlek ?

karlek:
@Frotty No, not until I get back to Sweden in late August.

Frotty:
To finally get this done I merged the two branches and fixed a bunch of issues such as binary expression formatting breaking parentheses and added support for differentiating between wurst, jurst and jass files. Then I implemented Jass formatting, which I find more useful because Wurst code is usually already formatted, while Jass code, e.g. from the war3map.j file, is not.
Some issues still remain with the Wurst formatting, which should be addressed in future PRs, but at least this can be merged now.

PR #640, opened 2018-04-12
1.29 compatability

Frotty:


de.peeeq.wurstscript/src/test/java/tests/wurstscript/tests/ClassesTests.java
```diff
@@ -1114,5 +1114,24 @@ public void tupleArrayMember() {
...
+                "   class Scouter",
+                "       static int size = 0",
+                "       construct()",
+                "           size++",
+                "",
+                "   init",
+                "       for i = 0 to 20000",
+                "           new Scouter()",
+                "       if Scouter.size == 20001",
+                "	        testSuccess()",
```
peq: Wouldn't it make more sense to create 20k powelevel objects instead of 20 scouters?

PR #651, opened 2018-05-09
[WIP] Branchmerger and Exitwhen optimizations

Frotty:
BranchMerger looks at the first two nodes of if branches in the controlflowgraph for every function.
If they are identical, and don't use variables that are affected by the if condition, these nodes can be merged and positioned before the if.

Nested Stacktraces majorly benefit from this (saving some 2k lines in EBR) because they have the same operations in different branches in a deterministic fashion.

The second optimization  merges consecutive exitwhen statements that wurst creates like so:
``` 
loop
    exitwhen i > 20
    exitwhen i < 0
```
will be rewritten to `exitwhen i > 20 or i < 0`

Also added simple commutative expression rewriting.


de.peeeq.wurstscript/src/main/java/de/peeeq/wurstscript/intermediatelang/optimizer/BranchMerger.java
```diff
@@ -0,0 +1,109 @@
...
+
+        // init in and out with empty sets
+        for (Node n : cfg.getNodes()) {
+            @Nullable ImStmt stmt = n.getStmt();
+            if (stmt != null && stmt.getParent() != null && stmt.getParent() instanceof ImIf) {
+                ImIf imIf = (ImIf) stmt.getParent();
+                if (n.getPredecessors().size() <= 1 && n.getSuccessors().size() == 2) {
+                    Node leftStmt = n.getSuccessors().get(0);
+                    Node rightStmt = n.getSuccessors().get(1);
+                    if (leftStmt.toString().equals(rightStmt.toString())) {
```
peq: Don't ever use `toString` in the compiler (unless you want to print something). What you probably want is ` if (leftStmt.getStmt().structuralEquals(rightStmt.getStmt())) {`

de.peeeq.wurstscript/src/main/java/de/peeeq/wurstscript/intermediatelang/optimizer/BranchMerger.java
```diff
@@ -0,0 +1,109 @@
...
+                        if (n.getPredecessors().size() == 1) {
+                            // Possible match. At last check if condition causes sideeffects.
+                            if (hasNoSideEffects(n, leftStmt)) {
+
+                                ImStmt mergedStmt = leftStmt.getStmt().copy();
```
peq: Not really necessary to copy here. You can probably just remove it from the old if-statement and add it to the new position.

de.peeeq.wurstscript/src/main/java/de/peeeq/wurstscript/intermediatelang/optimizer/BranchMerger.java
```diff
@@ -0,0 +1,109 @@
...
+
+    private List<ImVar> calculateVarUses(Node node) {
+        List<ImVar> result = Lists.newArrayList();
+        ImStmt stmt = node.getStmt();
+        if (stmt != null) {
+            stmt.accept(new ImStmt.DefaultVisitor() {
+                @Override
+                public void visit(ImFunctionCall call) {
+                    super.visit(call);
+                    visit(call.getFunc());
```
peq: To avoid endless loops you can use a similar approach like the StackTraceInjector:
```
Multimap<ImFunction, ImFunction> callRelationTr = Utils.transientClosure(callRelation);
```
This will give you a Multimap, where `callRelationTr.get(f))` will give you all functions called by `f` and also those called by functions called by `f` and so on (transitive closure).

With that it should be easier to calculate the set of variables that might be changed by each function.

de.peeeq.wurstscript/src/main/java/de/peeeq/wurstscript/intermediatelang/optimizer/BranchMerger.java
```diff
@@ -0,0 +1,109 @@
...
+                public void visit(ImVarArrayAccess va) {
+                    super.visit(va);
+                    result.add(va.getVar());
+                }
+
+                @Override
+                public void visit(ImSet va) {
+                    super.visit(va);
+                    result.add(va.getLeft());
+                }
```
peq: Missing some cases ... just search `jass_im.parseq` for `ref ImVar`.

de.peeeq.wurstscript/src/main/java/de/peeeq/wurstscript/intermediatelang/optimizer/BranchMerger.java
```diff
@@ -0,0 +1,109 @@
...
+        }
+        return result;
+    }
+
+    private boolean hasNoSideEffects(Node ifNode, Node stmtNode) {
+        List<ImVar> usesIf = calculateVarUses(ifNode);
+        List<ImVar> usesStmt = calculateVarUses(stmtNode);
+
+        usesIf.retainAll(usesStmt);
+        return usesIf.size() == 0;
```
peq: `hasNoSideEffects` should be named `sideEffectsCanAffectNode` or something like that: Or add a comment that you are checking, if executing `stmtNode` could affect the condition of the if-statement.

peq: I think you also have to include functions with side-effects. For example:

```
if (unit.getHp() > 100) {
   unit.setHp(50);
} else {
    unit.setHp(50);
}
```

de.peeeq.wurstscript/src/main/java/de/peeeq/wurstscript/intermediatelang/optimizer/SimpleRewrites.java
```diff
@@ -88,6 +94,11 @@ private void optimizeElement(Element elem) {
 
     }
 
+    private void optimizeConsecutiveExitWhen(ImExitwhen lookback, ImExitwhen element) {
+        lookback.getCondition().replaceBy(JassIm.ImOperatorCall(WurstOperator.OR, JassIm.ImExprs(lookback.getCondition().copy(), element.getCondition().copy())));
```
peq: copy can be avoided here, if you remove it from the current parent first

de.peeeq.wurstscript/src/main/java/de/peeeq/wurstscript/intermediatelang/optimizer/BranchMerger.java
Frotty: Doesn't work for me. This is false:

peq:
I changed some things (see commit messages).

Still have to see why build is failing, but otherwise looks ready to merge now.

PR #672, opened 2018-06-11
Experimental a + a = a * 2 rewrite

Frotty:
Not sure if it's too useful 😅


Frotty:
added a test and fix counting

de.peeeq.wurstscript/src/main/java/de/peeeq/wurstscript/intermediatelang/optimizer/SideEffectAnalyzer.java
```diff
@@ -219,4 +289,14 @@ public boolean cannotUseVar(ImStmt s, ImVar v) {
...
+
+    /**
+     * Checks if the given function calls any functions or modifies and variable
+     */
+    public boolean hasSideEffects(ImFunctionCall call) {
+        Set<ImFunction> natives = calledNatives(call.getFunc());
+        Set<ImFunction> directFuncs = calledFunctions(call.getFunc());
+        Set<ImVar> imVars = directlySetVariables(call.getFunc());
+        return natives.size() + directFuncs.size() + imVars.size() > 0;
+    }
```
peq: 1. You could define this function to take an arbitrary Element like the other functions, making it more general.
2. You could write this as `!calledNatives(call.getFunc()).isEmpty() && !calledFunctions(call.getFunc()).isEmpty() && !directlySetVariables(call.getFunc()).isEmpty()` that would make the computation a bit faster in some cases.

de.peeeq.wurstscript/src/main/java/de/peeeq/wurstscript/intermediatelang/optimizer/SimpleRewrites.java
```diff
@@ -189,6 +191,12 @@ private void optimizeOpCall(ImOperatorCall opc) {
                 } else {
                     wasViable = false;
                 }
+            } else if (left.structuralEquals(right) && opc.getOp() == WurstOperator.PLUS && !(left instanceof ImStringVal)) {
+                if (!(left instanceof ImFunctionCall && sideEffectAnalysis.hasSideEffects((ImFunctionCall) left))) {
```
peq: You should check for side-effects in general. Not only `ImFunctionCall`s can have side effects. There's also statement expressions, and expressions with side-effects nested in other expressios. For example `(1 + f()) + (1 + f())`.

PR #701, opened 2018-08-02
Get rid of empty init functions

Frotty:
Removes empty package init functions, see https://github.com/wurstscript/WurstScript/issues/511

### DIFF
--- a/de.peeeq.wurstscript/src/main/java/de/peeeq/wurstscript/translation/imoptimizer/InitFunctionCleaner.java
+++ b/de.peeeq.wurstscript/src/main/java/de/peeeq/wurstscript/translation/imoptimizer/InitFunctionCleaner.java
@@ -1,0 +1,59 @@
+package de.peeeq.wurstscript.translation.imoptimizer;
+
+import de.peeeq.wurstscript.jassIm.*;
+
+import java.util.Collection;
+import java.util.List;
+import java.util.Optional;
+import java.util.stream.Collectors;
+
+public class InitFunctionCleaner {
+
+    public static void clean(ImProg prog) {
+        List<ImFunction> initFuncs = prog.getFunctions().stream().filter(func -> func.getName().startsWith
+                ("init_") && func.getBody().size() == 1).collect(Collectors.toList());
+
+        initFuncs.forEach(func -> {
+            ImVar globalBridge = null;
+            for (ImVar global : prog.getGlobals()) {
+                if (global.getName().equals("ref_function_" + func.getName())) {
+                    globalBridge = global;
+                    break;
+                }
+            }
+            if (globalBridge != null) {
+                Collection<ImVarRead> imVarReads = globalBridge.attrReads();
+                if (imVarReads.size() == 1) {
+                    ImVarRead initRead = imVarReads.iterator().next();
+                    ImVar var = initRead.getVar();
+
+                    Optional<ImFunction> mainFunc = prog.getFunctions().stream().filter(f -> f.getName().equals("main")).findFirst();
+                    if (mainFunc.isPresent()) {
+                        ImStmts body = mainFunc.get().getBody();
+                        for (int i = 0; i < body.size(); i++) {
+                            ImStmt imStmt = body.get(i);
+                            if (imStmt instanceof ImFunctionCall) {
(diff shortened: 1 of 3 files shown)


de.peeeq.wurstscript/src/main/java/de/peeeq/wurstscript/translation/imoptimizer/InitFunctionCleaner.java
```diff
@@ -0,0 +1,59 @@
...
+public class InitFunctionCleaner {
+
+    public static void clean(ImProg prog) {
+        List<ImFunction> initFuncs = prog.getFunctions().stream().filter(func -> func.getName().startsWith
+                ("init_") && func.getBody().size() == 1).collect(Collectors.toList());
+
+        initFuncs.forEach(func -> {
+            ImVar globalBridge = null;
+            for (ImVar global : prog.getGlobals()) {
+                if (global.getName().equals("ref_function_" + func.getName())) {
```
peq: I don't like using the name here. Names should not matter in the intermediate language.

de.peeeq.wurstscript/src/main/java/de/peeeq/wurstscript/translation/imoptimizer/InitFunctionCleaner.java
```diff
@@ -0,0 +1,59 @@
...
+import java.util.Optional;
+import java.util.stream.Collectors;
+
+public class InitFunctionCleaner {
+
+    public static void clean(ImProg prog) {
+        List<ImFunction> initFuncs = prog.getFunctions().stream().filter(func -> func.getName().startsWith
+                ("init_") && func.getBody().size() == 1).collect(Collectors.toList());
+
+        initFuncs.forEach(func -> {
```
peq: please use normal for-loops -- better performance, better stack traces, better debugging, easier to read

de.peeeq.wurstscript/src/main/java/de/peeeq/wurstscript/translation/imoptimizer/InitFunctionCleaner.java
```diff
@@ -0,0 +1,59 @@
...
+                                    int finalI = i;
+                                    ((ImFunctionCall) imStmt).getArguments().forEach(arg -> {
+                                        if ((arg instanceof ImFunctionCall) && ((ImFunctionCall) arg).getFunc().getName().equals("Condition")) {
+                                            ImExpr next = ((ImFunctionCall) arg).getArguments().iterator().next();
+                                            if ((next instanceof ImVarAccess) && ((ImVarAccess) next).getVar().structuralEquals(var)) {
+                                                imStmt.replaceBy(JassIm.ImNull());
+                                                if (body.get(finalI - 1) instanceof ImFunctionCall) {
+                                                    body.get(finalI - 1).replaceBy(JassIm.ImNull());
+                                                }
+                                                body.get(finalI + 1).replaceBy(JassIm.ImNull());
```
peq: What is the -1 and +1 doing here? Looks like you are assuming that there are always certain other statements around a call to `TriggerAddCondition` but that does not seem very reliable to me.

de.peeeq.wurstscript/src/main/java/de/peeeq/wurstscript/translation/imoptimizer/InitFunctionCleaner.java
Frotty: What to use then?

de.peeeq.wurstscript/src/main/java/de/peeeq/wurstscript/translation/imoptimizer/InitFunctionCleaner.java
Frotty: The init calls are always generated the same way. Unless someone names a function "ref_function_" or so (which is the name issue above) then it seems reliable to me.

Frotty:
> Would it not be easier to improve the generation of the bridge-functions instead of trying to find and optimize them after the fact?

No idea where to start with that tbh. The old ticket & solution have been around for a while, this is one without inlining again.

Frotty:
@peq ?

PR #802, opened 2019-02-16
Fix compiletime implementation of FirstOfGroup

Nichts:
It no longer removes the unit from the group - just like it's jass counterpart

Minimal example:

```
function foo() returns bool
    let u = createUnit(players[0], 'hfoo', vec2(0, 0), angle(0))
    let g = CreateGroup()
    g.addUnit(u)
    FirstOfGroup(g)
    return not g.isEmpty()

@test function test_foo()
    assertTrue(foo())

init
    print(foo())
```

Used to print True in game, but the unit test failed


peq:
Thanks!

PR #803, opened 2019-02-17
Improve pjass error handling and add native osx executable

Nichts:
Tests now fail slightly more gracefully in case pjass can't be found or executed for some reason.

Also building the map on osx is now possible without wine or a vm (To actually test the map WC3 can be started from the command line - wc3libs sadly only supports windows).


Frotty:
Neat. I think the "equals" part might be a bit coarse and not detect all osx setups. Perhaps also use contains.
I already spoke with @WaterKnight and we will add osx support to wc3libs so you can run maps there as well.

Nichts:
Switched it to startsWith because that is what apache commons uses: (https://gitbox.apache.org/repos/asf?p=commons-lang.git;a=blob;f=src/main/java/org/apache/commons/lang3/SystemUtils.java;h=cc51f1196584b1cbd072ff9e663f8d6dd4214433;hb=HEAD)

Frotty:
Cool, thanks again for the PR. For integrity reasons I made a duplicate with an in house binary, see #806 
Closing this and merging the other. 👍

PR #824, opened 2019-03-25
WIP: Jhcr integration

peq:
Corresponding plugin update on branch https://github.com/wurstscript/wurst4vscode/tree/jhcr-commands

This adds 2 commands to vscode:

- Start a map with JHCR (will insert code for activating JHCR into the mapscript and runs `jhcr init` on the mapscript before injecting it into the map)
- Reload JHCR code (will rebuild the mapscript and run `jhcr update`)

TODO:

- [x] Make reload work
- [x] Make preload-path configurable (currently hardcoded for my machine)
- [x] Make jhcr executable path configurable (currently assumes it is on the path)
- [ ] Documentation
- [x] Remove line numbers from method names when translating closures (more stable identifiers)
- [ ] Maybe run inlining for global constants when updating mapscript

### DIFF
--- a/de.peeeq.wurstscript/parserspec/wurstscript.parseq
+++ b/de.peeeq.wurstscript/parserspec/wurstscript.parseq
@@ -313,6 +313,7 @@ Modifier =
 HasModifier = NameDef | TypeDef | ModuleDef | ConstructorDef | GlobalVarDef | FunctionDefinition
 HasTypeArgs =  ExprNewObject | FunctionCall | ModuleUse | StmtCall | TypeExprSimple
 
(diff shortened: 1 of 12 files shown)


lep:
It could be a bug in jhcr. You are basically creating over 32k new instructions which is possible since old ones currently are not being free'd. Best way to debug would be posting the output of `jhcr update`: that should show for how many functions are being updated. I guess passing the `--asm` flag to update should also give an overview of how many instructions are being inserted. Every line in the asm output correspondens to one instruction. 

And it would be nice to know if you hit that error with only one update/reload or multiple ones.

PR #847, opened 2019-05-06
added new jasstypes for ptr 1.31

Krypt0n:


Frotty:
Hm, I think codebeat check is broken oO.
Looking good. I think we can merge this before 1.31 hits main as it's just for highlight and name restriction iirc.

Frotty:
Totally forgot about this, sorry 😅 
Thanks 👍

PR #848, opened 2019-05-10
Implement BlzGroupUnitAt, BlzGroupAddGroupFast, BlzGroupRemoveGroupFa…

Krypt0n:
…st for compiletime

### DIFF
--- a/de.peeeq.wurstscript/src/main/java/de/peeeq/wurstio/jassinterpreter/providers/GroupProvider.java
+++ b/de.peeeq.wurstscript/src/main/java/de/peeeq/wurstio/jassinterpreter/providers/GroupProvider.java
@@ -74,4 +74,35 @@ public class GroupProvider extends Provider {
         LinkedHashSet<IlConstHandle> groupList = (LinkedHashSet<IlConstHandle>) group.getObj();
         return ILconstInt.create(groupList.size());
     }
+	
+    public ILconst BlzGroupUnitAt(IlConstHandle group, ILconstInt index) {
+        LinkedHashSet<IlConstHandle> groupList = (LinkedHashSet<IlConstHandle>) group.getObj();
+        ILconst elem = ILconstNull.instance();
+        if(index.getVal() > 0) {
+            for(int i = 0; i < index.getVal(); i++)
+                elem = groupList.iterator().next();
+            return elem;
+        }
+        else if(index.getVal() == 0)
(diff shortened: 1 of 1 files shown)


peq:
Thanks!

de.peeeq.wurstscript/src/main/java/de/peeeq/wurstio/jassinterpreter/providers/GroupProvider.java
```diff
@@ -74,4 +74,35 @@ public ILconstInt BlzGroupGetSize(IlConstHandle group) {
         LinkedHashSet<IlConstHandle> groupList = (LinkedHashSet<IlConstHandle>) group.getObj();
         return ILconstInt.create(groupList.size());
     }
+	
+    public ILconst BlzGroupUnitAt(IlConstHandle group, ILconstInt index) {
+        LinkedHashSet<IlConstHandle> groupList = (LinkedHashSet<IlConstHandle>) group.getObj();
+        ILconst elem = ILconstNull.instance();
+        if(index.getVal() > 0 && index.getVal() < groupList.size()) {
+            for(int i = 0; i < index.getVal(); i++)
+                elem = groupList.iterator().next();
```
peq: This looks like it would always  return the first element of the group, since you call `iterator` inside the loop. Maybe you can do another pull request to fix this?

de.peeeq.wurstscript/src/main/java/de/peeeq/wurstio/jassinterpreter/providers/GroupProvider.java
```diff
@@ -74,4 +74,35 @@ public ILconstInt BlzGroupGetSize(IlConstHandle group) {
...
+            return FirstOfGroup(group);
+        return elem;
+    }
+    
+    public ILconstInt BlzGroupAddGroupFast(IlConstHandle group, IlConstHandle groupAdd) {
+        LinkedHashSet<IlConstHandle> groupList    = (LinkedHashSet<IlConstHandle>) group.getObj();
+        LinkedHashSet<IlConstHandle> groupListAdd = (LinkedHashSet<IlConstHandle>) groupAdd.getObj();
+        groupListAdd.forEach((IlConstHandle u) -> {
+            groupList.add(u);
+        });
```
peq: Could probably use `addAll` instead.

PR #867, opened 2019-06-21
Second try to add Lua support

peq:
Work-in-progress branch for Lua support.


jlfarris91:
What is the status of the lua backend? This pr makes it look like it's already in. If so, is there any documentation on how to enable it?

Frotty:
It is in, just add `-lua` to your wurst_run.args file.
But it still has many bugs as you can see from the open tickets.

PR #875, opened 2019-07-12
Custom wc3 run arguments

WareditorGit:
Added a way to set custom wc3 run arguments for the run map command via vscode settings.
If this merged, should also update the vscode plugin.


Frotty:
cool 👍

PR #893, opened 2019-09-27
Permit trivial TRVE usage in jass input

Frotty:
did this for w3p, should work?
Needs some fixing and a test for compiler use.


de.peeeq.wurstscript/src/main/java/de/peeeq/wurstscript/validation/TRVEHelper.java
```diff
@@ -0,0 +1,7 @@
+package de.peeeq.wurstscript.validation;
+
+import java.util.HashSet;
+
+public class TRVEHelper {
+    public static final HashSet<String> TO_KEEP = new HashSet<>();
```
peq: Please don't use mutable static variables. I would prefer to have this as a variable in `ImTranslator`. 

Also the all caps style is usually only used for immutable constants.

de.peeeq.wurstscript/src/main/java/de/peeeq/wurstscript/validation/WurstValidator.java
```diff
@@ -1958,11 +1958,13 @@ private void checkForDuplicatePackages(WurstModel model) {
     }
 
     private void checkBannedFunctions(ExprFunctionCall e) {
-        String[] banned = new String[]{
-                "TriggerRegisterVariableEvent" /* , "ExecuteFunc" */};
-        for (String name : banned) {
-            if (e.getFuncName().equals(name)) {
-                e.addError("The function " + name + " is not allowed in Wurst.");
+        if (e.getFuncName().equals("TriggerRegisterVariableEvent")) {
+            if (e.getArgs().get(1) instanceof ExprStringVal) {
```
peq: Maybe add a check `e.getArgs().size() > 1` to avoid out of bounds exception.

de.peeeq.wurstscript/src/main/java/de/peeeq/wurstscript/validation/TRVEHelper.java
Frotty: it doesn't work as variable in ImTranslator because I need to add names from the validator.

Frotty:
Okay I added 2 jass unit tests and did some real world map tests.
This works for direct usage or simply wrapped (hooked) usage.
And yes, the point of this is mainly to support legacy/vJass scripts, not to allow it in wurst.

Frotty:
I have been using this impl in w3protect for some time - so far without issues. I think we should only allow it in Jass though, to prevent the issues discussed.

Frotty:
@peq anything stopping this from being merged?

de.peeeq.wurstscript/src/main/java/de/peeeq/wurstio/jassinterpreter/providers/TriggerProvider.java
```diff
@@ -47,6 +44,12 @@ public void TriggerRegisterPlayerChatEvent(IlConstHandle trigger, IlConstHandle
 //        triggerMock.registerEvent();
     }
 
+    public void TriggerRegisterVariableEvent(IlConstHandle trigger, ILconstString varName, IlConstHandle opcode, ILconstReal limitval) {
+        TriggerMock triggerMock = (TriggerMock) trigger.getObj();
+        // TODO
+//        triggerMock.registerEvent();
```
peq: I guess this should just be an empty mock, so remove the Todo?

de.peeeq.wurstscript/src/main/java/de/peeeq/wurstscript/translation/imoptimizer/ImOptimizer.java
```diff
@@ -118,12 +119,12 @@ public void visit(ImSet e) {
...
                         if (e.getLeft() instanceof ImVarAccess) {
                             ImVarAccess va = (ImVarAccess) e.getLeft();
-                            if (!trans.getReadVariables().contains(va.getVar())) {
+                            if (!trans.getReadVariables().contains(va.getVar()) && !TRVEHelper.protectedVariables.contains(va.getVar().getName())) {
                                 replacements.add(Pair.create(e, Collections.singletonList(e.getRight())));
                             }
                         } else if (e.getLeft() instanceof ImVarArrayAccess) {
                             ImVarArrayAccess va = (ImVarArrayAccess) e.getLeft();
-                            if (!trans.getReadVariables().contains(va.getVar())) {
+                            if (!trans.getReadVariables().contains(va.getVar()) && !TRVEHelper.protectedVariables.contains(va.getVar().getName())) {
```
peq: Would it be possible to just count TRVE as a read?

de.peeeq.wurstscript/src/main/java/de/peeeq/wurstscript/validation/TRVEHelper.java
```diff
@@ -0,0 +1,7 @@
+package de.peeeq.wurstscript.validation;
+
+import java.util.HashSet;
+
+public class TRVEHelper {
+    public static final HashSet<String> protectedVariables = new HashSet<>();
```
peq: Please don't use static variables with mutable state.

de.peeeq.wurstscript/src/main/java/de/peeeq/wurstscript/validation/WurstValidator.java
```diff
@@ -1960,11 +1979,33 @@ private void checkForDuplicatePackages(WurstModel model) {
...
-                e.addError("The function " + name + " is not allowed in Wurst.");
+        if (e.getFuncName().equals("TriggerRegisterVariableEvent")) {
+            if (e.getArgs().size() > 1) {
+                if (e.getArgs().get(1) instanceof ExprStringVal) {
+                    ExprStringVal varName = (ExprStringVal) e.getArgs().get(1);
+                    TRVEHelper.protectedVariables.add(varName.getValS());
+                    WLogger.info("keep: " + varName.getValS());
+                    return;
+                } else if (e.getArgs().get(1) instanceof ExprVarAccess) {
+                    // Check if this is a two line hook... thanks Bribe
```
peq: Is this for a specific bj function? Maybe add a comment...

Frotty: This is for single nested calls iirc to detect them as statically resolvable. bribe used to use it in his damage engine.
```
function myfunc (with same params and return typ)
    TriggerRegisterVariableEvent(...)
```
A bit bad solution maybe :D

de.peeeq.wurstscript/src/main/java/de/peeeq/wurstscript/translation/imoptimizer/ImOptimizer.java
Frotty: how?

de.peeeq.wurstscript/src/main/java/de/peeeq/wurstscript/validation/TRVEHelper.java
Frotty: You didn't answer last time. I can't add it as variable to the ImTranslator because I need it in the validator. You want it in the outer scope of that?

PR #894, opened 2019-10-10
added hidden natives for BlzGetTriggerFrameValue, BlzGetTriggerFrameText

Krypt0n:


Krypt0n:
denied, do changes locally if needed, might be updated to official common.j by blizz

PR #907, opened 2019-11-22
Add option for disabling map configuration when building or running maps

Thorminator:
Option for disabling map configuration.

Useful workaround when running and building reforged maps, since their W3I format is not yet supported.


Frotty:
THe changes themselves look okay, but it doesn't seem like the best solution for me.
Probably instead the exception should just be caught and map config not applied when it can't, for whatever unknown errors.
I believe most ppl encountering the issue would not find this config option.

Thorminator:
That might be a better approach, I'll rewrite it and open a new PR =)

PR #922, opened 2020-01-14
Show warning if constant variable is not defined using 'let'

Frotty:
See https://github.com/wurstscript/WurstScript/issues/817
Stdlib has been adjusted.


Frotty:
> Why are you checking for ExprMemberVar if this is only checking local variables?

That's for the left part of an assignment. Tuples would be detected as unused otherwise.

Issue #679, opened 2018-06-21
New Generics


peq:
Currently, generics are translated using erasure. This means, that generics can only be instantiated with types compatible with `int`. The idea of this proposal is to change the translation and instead of using erasure translation would generate one copy per instantiated type. This would allow to extend generics with some interesting new features, but would also be incompatible in some cases.

Language changes:
============

## 1. Allow all types for generics

Every type can be inserted for a generic type parameter without any restrictions.

For example, we can then use `LinkedList<vec3>`.

This also allows us to get rid of the implicit calls to `fromIndex` and `toIndex` that we currently have as a woraround.

## 2. Allow to restrict type parameters / type classes

Sometimes we want to restrict type parameters.
For example `LinkedList<T>.toString() returns string` only can be implemented, if T has a `toString` method.

This is comparable to [statically resolved type parameters in F#](https://docs.microsoft.com/en-us/dotnet/fsharp/language-reference/generics/statically-resolved-type-parameters), type classes in Haskell and a bit like type bounds in Java.

Syntactically this can look somewhat like this:

```
typeclass Show
    function toString() returns string

function LinkedList<T>.toString<T : Show>() returns string
```

```
typeclass Serializable
    function encode() returns string
    static function decode(string s) returns thistype

function LinkedList<T>.encode<T : Serializable>() returns string
function LinkedList<T>.decode<T : Serializable>() returns string
```

Type classes would be instantiated implicitly (it is enough to implement the required methods). We would also extend extension functions so that new static methods can be implemented for a type.
If there are multiple fitting methods in scope, it is a compiletime error. The instantiation is part of the type, so the following example is not allowed:

```
package A

public typeclass Ord
    function lessEq(thistype other) returns boolean

function string.lessEq(string other) returns boolean
    return this.length() <= other.length()

public class TreeSet<T : Ord>
    ...

public TreeSet<string> mySet = TreeSet.create("a","aa","aaa")

package B
import A

function string.lessEq(string other) returns boolean
    return /* some other order */

init
    TreeSet<string> s = mySet // error, because Ord intance / lessEq is not compatible


```

## 3. Disallow casting to int

Since we now allow all types to be used for type parameters, we can no longer cast type parameters to `int`. However, we can use the new feature of restricted type parameters to implement the same feature:

Old `HashMap`:

```
public class HashMap<K,V> extends Table

	/** Whether a value exists under the given key or not */
	function has(K key) returns boolean
		return hasInt(key castTo int)

[...]

PR #931, opened 2020-01-31
[WIP] Type classes

peq:
This branch contains an implementation of type classes as described in the new generics design in #679.

TODO list:

- [x] Instance Declarations
- [x] Basic Type Bounds
- [x] Type bounds in classes
- [x] Dependent type classes
- [x] Efficient search + namespacing (only search for instances in imported packages)
- [x] implicit instances for classes: ToIndex, FromIndex (maybe: TableKey, TableValue, Hashable)
- [x] Type Check: All functions implemented correctly
- [x] Add check for diverging instance search
- [x] Also search for supertypes
- [x] Also search for methods in extended interfaces
- [x] Overloads with extension methods / type classes should not give error.

Maybe add:

- [ ] Complex inference (see Iterable example)
- [ ] Translate casting to int using ToIndex and FromIndex type classes (better backwards compatibility)
- [ ] Use type classes for iterators (for-loops)
- [ ] Use type classes for operator overloading 
- [ ] Add an expression (or magic function) for obtaining type class instances and calling methods on them, like (implicitly<Show<int>>.toString(42))
- [ ] Automatically use existing methods and extension functions in instance declarations


Frotty:
Nice 👍 
I'm not too sure about the name `instance` since we frequently use it in the documentation to refer to a class instance.

peq:
Yeah, will probably change it as it also introduces a new keyword and thus breaks StdLib tests.

Maybe just `implements Show<int>` or `for <T: Show> implements Show<List<T>>`?

Frotty:
I get this error when trying to compile forest defense:

PR #932, opened 2020-01-31
Allow equality checks between agent types

Frotty:
Jass allows this 🤷‍♂


Frotty:
Done. There are now 2 separate tests.

PR #937, opened 2020-02-10
Support 1.32 on linux

Frotty:


Frotty:
not needed anymore with latest patch

Frotty:
Dunno why not merged yet

PR #951, opened 2020-05-11
reforged

Cokemonkey11:
😱 

In updating w3libs for reforged I took a look at how WurstScript uses it, and wasn't quite satisfied that updating w3libs would be sufficient to improve stability.

This PR bundles a few things all in one 💩:

- w3libs update (better handling of wc3 installations, both reforged and for backwards compat)
- refactoring W3utils (no longer static; should be easier to reason about)
- removed dependency on shell-exec for `git` (wasn't working for me in vscode so I rewrote to use jgit)
- some cleanup (dead code, commented-out code, 120-column line limit
- reworked some key workflows to use `Optional` instead of @Nullable and @NotNulll


Frotty:
> This PR bundles a few things all in one

Why? just makes review more difficult. way too many changes.

de.peeeq.wurstscript/src/main/java/de/peeeq/wurstio/languageserver/requests/CodeActionRequest.java
```diff
@@ -61,32 +62,38 @@ public CodeActionRequest(CodeActionParams params, BufferManager bufferManager) {
...
         // get element under cursor
-        Element e = Utils.getAstElementAtPos(cu, line, column, false);
+        Optional<Element> e = Utils.getAstElementAtPos(cu, line, column, false);
 
 
         WLogger.info("Code action on element " + Utils.printElementWithSource(e));
-        if (e instanceof ExprNewObject) {
-            ExprNewObject enew = (ExprNewObject) e;
+        if (e.isEmpty()) {
+            // TODO non simple TypeRef
```
Cokemonkey11, line 70: Where did this line come from?

de.peeeq.wurstscript/src/main/java/de/peeeq/wurstio/languageserver/requests/Colors.java
```diff
@@ -100,18 +101,24 @@ public ColorPresentationRequest(ColorPresentationParams params) {
...
+            Optional<Element> elemAtPos = Utils.getAstElementAtPos(cu.get(), range.getStart().getLine() + 1,
+                range.getStart().getCharacter() + 2, false);
+            if (elemAtPos.isEmpty()) {
+                return Collections.emptyList();
+            }
+            Element elem = elemAtPos.get();
 
             if (elem instanceof ExprIntVal
-                    && elem.getParent() != null) {
+                    && Optional.ofNullable(elem.getParent()).isPresent()) {
```
Cokemonkey11, line 117: @peq I see that Wscope.java is AST-generated. Is there any way to make it return Optional for getParent?

de.peeeq.wurstscript/src/main/java/de/peeeq/wurstio/languageserver/requests/CodeActionRequest.java
Cokemonkey11: Ah it's copy-pasted from the final return in this function. Unclear to me if the TODO is relevant here as well.

Cokemonkey11:
@peq @Frotty I see that the build failed because `Optional::isEmpty` was added in Java 11.

Is an older java compat needed, or is it possible to upgrade?

Cokemonkey11:
> 
> 
> > This PR bundles a few things all in one
> 
> Why? just makes review more difficult. way too many changes.

Hey @Frotty thanks for the comment. I agree, it's not good that it ended up this way. I was going down a rabbit hole and elected not to think about it much and just kept digging.

I am hoping that it won't be too contentious since commit volume on the compiler is relatively low.

Frotty:
yes, target is 8.

Cokemonkey11:
Thanks @Frotty 

I didn't realise that Java works in a way that sourceCompatibility is ignored during compile time. I will push a commit to fix this shortly.

de.peeeq.wurstscript/src/main/java/de/peeeq/wurstio/languageserver/requests/RunTests.java
```diff
@@ -366,7 +369,7 @@ public void showInfoMessage(String message) {
     }
 
     private class TestTimeOutException extends Throwable {
-
+        private static final long serialVersionUID = -7080085729129882874L;
```
Frotty, line 372: ?!

de.peeeq.wurstscript/src/main/java/de/peeeq/wurstio/utils/W3Utils.java
```diff
@@ -6,77 +6,90 @@
...
 import net.moonlightflower.wc3libs.port.StdGameExeFinder;
-import net.moonlightflower.wc3libs.port.mac.MacGameVersionFinder;
+import net.moonlightflower.wc3libs.port.StdGameVersionFinder;
 import net.moonlightflower.wc3libs.port.win.WinGameExeFinder;
 
 import java.io.File;
 import java.io.IOException;
+import java.util.Optional;
 
 public class W3Utils {
```
Frotty, line 16: Is there any circumstance where this class not being a singleton is required?

de.peeeq.wurstscript/src/main/java/de/peeeq/wurstio/languageserver/requests/MapRequest.java
```diff
@@ -344,37 +348,43 @@ protected File compileScript(WurstGui gui, ModelManager modelManager, List<Strin
...
             println("We will try to start the map now, but it will probably fail. ");
         }
         return compiledScript;
     }
 
-    private void parseCustomPatchVersion() {
-        if (wc3Path != null) {
-            W3Utils.parsePatchVersion(new File(wc3Path));
-            if (W3Utils.getWc3PatchVersion() == null) {
+    private W3Utils parseCustomPatchVersion() throws RequestFailedException {
```
Frotty, line 378: if `W3Utils` is not gonna be singleton, then rename it to something proper (e.g. W3InstallationData or so) and probably also rename the method

Frotty:
I guess this could be nice, but it really contains too many irrelevant changes for the actual content to be properly reviewed.

de.peeeq.wurstscript/src/main/java/de/peeeq/wurstio/languageserver/requests/RunTests.java
Cokemonkey11: It's a generated value for a serialisable class. IDE flagged it, I'm not sure if it's actually needed or not

de.peeeq.wurstscript/src/main/java/de/peeeq/wurstio/utils/W3Utils.java
Cokemonkey11: It wasn't a singleton before, it was static

Cokemonkey11:
I disagree, I reviewed it myself in one evening. There is a lot of repetition and the interesting content is sparse

I suggest to use this tooling to make it more digestible:

image

de.peeeq.wurstscript/src/main/java/de/peeeq/wurstio/languageserver/requests/RunTests.java
Frotty: > It's a generated value for a serialisable class

Is this class serialized?

Cokemonkey11: I don't know sorry, but my language server flagged it, so maybe yes?

de.peeeq.wurstscript/src/main/java/de/peeeq/wurstio/languageserver/requests/MapRequest.java
Cokemonkey11: Thanks! Fixed

Cokemonkey11:
Hey, I didn't see any comments here in a while. It seems that review is stalling here because it's not clear who is expected to review and also because of the burden of high number of changes.

I can do some surgery and PR one change at a time, but before spending time on that it would be good to know what changes people are on-board with and if it will even get a review on that form.

@Frotty @peq are you prepared to spend time on reviewing? Am I right that splitting this up will get 👀 sooner? Are there any changes here that you're not on-board with?

I do not think this should be needed tbh. You should just review this PR - it's not even 1000 lines.

Frotty:
> it's not even 1000 lines.

True, then it should be no problem for you to separate features from refactorings 👍

Cokemonkey11:
No, it would be quite painful. These changes are not purely _orthogonal_.

I think the onus is on you the maintainers to at least review this at a high level. I have provided bullet points of the 5 main changes.

Frotty:
I made the missing changes on  another branch because github desktop didn't let me push to this one.
see https://github.com/wurstscript/WurstScript/pull/958

PR #957, opened 2020-06-18
adding support for importing *.dds files

Krypt0n:


Frotty:
Not really useful unless you re-generate the stdlib asset packages and merge additions I believe?

Krypt0n:
Well in future dds will replace any blp most likely. And well only use dds right now due to 4x/6x compression, depends on with or without alpha. Not sure when, wc3 is a messy mysterium atm.

PR #962, opened 2020-07-28
clarify how versions are changed

Cokemonkey11:


Frotty:
I doesn't seem evident to me why you replace the manual deploy instructions with the jenkins explanation?

Cokemonkey11:
@Frotty I'm happy to add a new line and/or undelete the following line:

```
 - Run gradle task `:versionInfoFile` 
```

In any case, the motivation here was that I realised that it isn't discoverable the mistake I made in the `reforged` PR which changes tag behavior. Probably no one will make the same mistake, but since this doc is clarifying some dev process, it seems worthwhile.

What do you advise?

PR #963, opened 2020-07-30
Fix tags

Cokemonkey11:
- Reintroduces the jgit change applied with !951 and !958
- Fix the regression in writing git tag in CompileTime.java by reading the tag
- tested that this behaves correctly with tag and does something not crazy without tag

### DIFF
--- a/de.peeeq.wurstscript/build.gradle
+++ b/de.peeeq.wurstscript/build.gradle
@@ -5,6 +5,7 @@ buildscript {
     dependencies {
         classpath 'de.undercouch:gradle-download-task:3.2.0'
         classpath 'org.kt3k.gradle.plugin:coveralls-gradle-plugin:1.0.2'
+        classpath 'org.eclipse.jgit:org.eclipse.jgit:5.7.+'
     }
 }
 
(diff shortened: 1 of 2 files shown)


Frotty:
need test/proof that this works before merge.

Cokemonkey11:
Hey @Frotty thanks for taking a look.

I tested this manually, but didn't add any unit tests:

```bash
de.peeeq.wurstscript $ git tag
test
de.peeeq.wurstscript $ ./gradlew :versionInfoFile

Deprecated Gradle features were used in this build, making it incompatible with Gradle 6.0.
Use '--warning-mode all' to show the individual deprecation warnings.
See https://docs.gradle.org/5.0/userguide/command_line_interface.html#sec:command_line_warnings

BUILD SUCCESSFUL in 1s
1 actionable task: 1 executed
de.peeeq.wurstscript $ find . -name 'CompileTimeInfo.java' | xargs cat | rg version=
            public static final String version="1.8.1.0-test";
```

Is that acceptable?

Frotty:
i dont even know what u mean by tag. The tasks reads the commit hash and adds it to the version so when someone pastes a log u know they have a custom build, and you know which commit it is from.

Cokemonkey11:
@Frotty you're mistaken. The commit hash and custom builds is not all it does. Hudson reads the tag from the git repo - i.e. CI depends on this behavior - see http://peeeq.de/hudson/job/Wurst/1257/consoleFull

You'll recall that the reason #951 was partially reverted is because the move to jgit broke CI.

This PR is to fix that.

Btw, all of this is explained in the original comment from 30th July

Issue #984, opened 2021-02-05
Only 1 modified default object is built


jmrico01:
**Describe the bug**
My main Windows PC has `Installed Compiler Build: 1259`, and my Windows laptop has `Installed Compiler Build: 1262`, which is the latest version. 1259 builds the map as expected - all the custom objects and modified defaults are added to the map, and I can see them in the `objectEditingOutput/WurstExportedObjects_w*a.wurst.txt` files. For the exact same project, 1262 builds the custom objects just fine, but only includes 1 modified default object for all object categories. This is reflected both in the `war3map.w3*` binaries in the map MPQ and in the `objectEditingOutput/WurstExportedObjects_w*a.wurst.txt` files.

I'm attaching the built map and `objectEditingOutput` for each version, as well as the `compiler.log` file. `languageServer.log` was empty for both compiler versions.
- [build-1259.zip](https://github.com/wurstscript/WurstScript/files/5930495/build-1259.zip)
- [build-1262.zip](https://github.com/wurstscript/WurstScript/files/5930516/build-1262.zip)

If the issue isn't too obvious and you need to actually reproduce the build, I can give you access to the project. Just don't have too much time atm. Here's the base map I'm using, for good measure:
[base.zip](https://github.com/wurstscript/WurstScript/files/5930532/base.zip)

**EDIT:** I tested a build with 1262 using only the `base.w3x` and a very vanilla `war3map.j`, no custom objects at all, and the issue still happens. Should be easy to reproduce on your end.

Frotty:
probably aadcf2905d4fab61ee6c05a010a3393cb29fc86c is the cause then, but I don't see exactly why.

jmrico01:
maybe this? idk! https://github.com/wurstscript/WurstScript/commit/aadcf2905d4fab61ee6c05a010a3393cb29fc86c#diff-a431aa1b9398e51369f3b3d4418b9f3c1804c503269b3a3f6f6e8f936e57bf67L18

PR #985, opened 2021-02-06
Use original object ID if new object ID is 0

jmrico01:
Simplest fix for #984, based on [these few comments](https://github.com/wurstscript/WurstScript/commit/aadcf2905d4fab61ee6c05a010a3393cb29fc86c#diff-a431aa1b9398e51369f3b3d4418b9f3c1804c503269b3a3f6f6e8f936e57bf67L18). If this is all that's needed, great! If you had other plans for this fix or wanted to include some more changes, please feel free to close this PR.

Also changed `Map` to `LinkedHashMap` per @peq's comment, to keep the order deterministic.


Frotty:
Nice. Could only become a problem if someone assigns 0 als newId but I guess that's fine. 👍

Issue #885, opened 2019-08-13
Ingame errors with -lua


Jampi0n:
I tried the -lua run argument in my map and it produced errors during the initialization. One of the issues I found is, that functions can return nil instead of the specified type, possibly leading to errors, when using the returned value.

code to reproduce: https://bin.wurstlang.org/ajosinetib.wurst

Could be related to #869

Jampi0n:
I managed to change my map, so it works in lua. The most prevalent issue is mentioned above. In particular this construct does not work:
let tmp = HashMap.get(key)
if tmp != null
This can be fixed using HashMap.has(key) before.

Additionally using null as string does not work.

For the first issue, there is likely no good solution, other than updating the code and check whether the key exists first.
I assume null strings can work by simply replacing null with "", when compiling to lua (is there a difference between null and "" in jass?)

The standard library uses the HashMap as shown above, so it can lead to issues. It also uses null as string, so at the moment FrameHandle.wurst cannot initialize.

peq:
Thanks for the report.

The thing with nulls and strings is something I thought was working. We have some test cases testing different combinations of using `null` here: https://github.com/wurstscript/WurstScript/blob/d79f738cfee4d10e288097755d17b165985c3c39/de.peeeq.wurstscript/src/test/java/tests/wurstscript/tests/BugTests.java#L1083-L1123

Can you show an example where null as string does not work?

I'll also look at the HashMap problem. Maybe that will already fixed by the upcoming changes in #931 .

Jampi0n:
Using the FrameHandle package leads to an error, because it uses
`let mouseCage = createFrame("FRAME", "SetMousePositionCage", GAME_UI, null, 0)`
in line 389. Here, null is used as parameter for a native function that expects a string.

So using that package, you get an error with the string. For example with this code: https://bin.wurstlang.org/viyirecelo.wurst

On a related note, I found an issue when I tried to create a minimal example with the string problem.
If I add the imports NoWurst and Framehandle to minimize the produced code, the map doesn't start in lua mode (JASS works fine). I only get the screen with the lordaeron lion. Code: https://bin.wurstlang.org/epuhotezaj.wurst

Frotty:
seems resolved.

PR #995, opened 2021-04-11
fix lua null string

Jampi0n:
replaces null strings with empty strings in lua

Null strings will cause errors when used in lua natives.
Fixes one of the issues reported in #885 .

### DIFF
--- a/de.peeeq.wurstscript/src/main/java/de/peeeq/wurstscript/translation/lua/translation/ExprTranslation.java
+++ b/de.peeeq.wurstscript/src/main/java/de/peeeq/wurstscript/translation/lua/translation/ExprTranslation.java
@@ -102,7 +102,11 @@ public class ExprTranslation {
     }
 
     public static LuaExpr translate(ImNull e, LuaTranslator tr) {
-        return LuaAst.LuaExprNull();
+        if(isStringType(e.getType())) {
+            return LuaAst.LuaExprStringVal("");
+        } else {
+            return LuaAst.LuaExprNull();
+        }
     }
 
     public static LuaExpr translate(ImOperatorCall e, LuaTranslator tr) {
--- a/de.peeeq.wurstscript/src/test/java/tests/wurstscript/tests/LuaTranslationTests.java
+++ b/de.peeeq.wurstscript/src/test/java/tests/wurstscript/tests/LuaTranslationTests.java
@@ -1,0 +1,132 @@
+package tests.wurstscript.tests;
+
+import com.google.common.base.Charsets;
+import com.google.common.io.Files;
+import org.testng.annotations.Test;
+
+import java.io.File;
+import java.io.IOException;
+
+import static org.testng.AssertJUnit.*;
+
+
+public class LuaTranslationTests extends WurstScriptTest {
+
+
+    @Test
+    public void testStdLib() throws IOException {
+        test().testLua(true).withStdLib().lines(
+            "package Test",
+            "import MagicFunctions",
+            "init",
+            "   print(compiletime)"
+        );
+        String compiled = Files.toString(new File("test-output/lua/LuaTranslationTests_testStdLib.lua"), Charsets.UTF_8);
+        assertTrue(compiled.contains("MagicFunctions_compiletime"));
+    }
+
+    @Test
+    public void testExecution() {
+        test().testLua(true).executeProg().lines(
+            "package Test",
+            "native testSuccess()",
+            "init",
+            "   testSuccess()"
+        );
+    }
+
+    @Test
+    public void nullString1() throws IOException {
+        test().testLua(true).lines(
+            "package Test",
+            "function nullString() returns string",
+            "	return null",
+            "init",
+            "   nullString()"
+        );
+        String compiled = Files.toString(new File("test-output/lua/LuaTranslationTests_nullString1.lua"), Charsets.UTF_8);
+        assertTrue(compiled.contains("return \"\"") && !compiled.contains("return nil"));
+    }
+
+    @Test
+    public void nullString2() throws IOException {
+        test().testLua(true).lines(
+            "package Test",
+            "function takesString(string s)",
+            "init",
+            "   takesString(null)"
+        );
+        String compiled = Files.toString(new File("test-output/lua/LuaTranslationTests_nullString2.lua"), Charsets.UTF_8);
+        assertTrue(compiled.contains("takesString(\"\")") && !compiled.contains("takesString(nil)"));
+    }
+
+    @Test
+    public void nullString3() {
+        test().testLua(true).executeProg().lines(
+            "package Test",
+            "native testSuccess()",
+            "function nullString() returns string",
+            "   return null",
+            "function returnsString(string s) returns string",
+            "   return s + nullString() + s",
+            "init",
+            "   let s = \".\"",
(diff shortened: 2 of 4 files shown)


Frotty:
Hey, thanks for the lua PRs. Could you add unit tests for them?

Jampi0n:
How would these unit tests look like?
Most of the unit tests I've seen are wurst programs which are expected to either be valid wurst code or not (testAssertOkLines, testAssertErrorsLines).
In this case the translation to lua needs to be tested.

Frotty:
Yes, tests output the output program states into files which you parse (test-output).
Some already do this, see optimizer correctness tests. Dunno if a lua one exists, but the test should generate a lua output.

Jampi0n:
I have some test cases, but I had to change WurstScriptTest.java to get lua compilation to work. In line [246](https://github.com/wurstscript/WurstScript/blob/4cfa7f69a61df68f8815bcfebc10377ee43930f2/de.peeeq.wurstscript/src/test/java/tests/wurstscript/tests/WurstScriptTest.java#L246).

Is there a reason why lua is only tested if executeProg is true? executeProg is a parameter for translateAndTestLua, so it seems unnecessary to put it in the if. It would also be nice to have withStdLib for lua as well.

I removed executeProg from the if and the tests work now, but I don't want to commit it, because there are probably reasons why it was put like this.

Frotty:
It seems fine to me since executeProg is handled in the lua compile method.
Stdlib is deactivated, because it didn't (doesn't ?) compile with lua I suppose.
Both of these seem like left overs from early development and can probably be removed, ideally add more tests for this as well.

Jampi0n:
I changed the if to only check for lua.
Executing lua works fine, though it doesn't work with most natives from the standard library. E.g. InitHashtable or CreateTrigger doesn't work.
Standard library by itself also works.

I added tests to verify that both lua execution and using standard library work.
For the null string issue I added tests for strings and other nullable types to make sure only the string nulls are replaced by "".

Frotty:
Nice, sounds goo so far.
Please resolve conflicts with master and make pipeline passable. Perhaps can use some different docker image?

Jampi0n:
I don't know much about docker, so I guess just remove the tests for now. Should I leave the functions in and just remove the test annotation or remove the function as well?

Are you referring to the hashtable issue in #885 ?
The issue seems to be much more limited than I initially thought. It's not that hashtables return nil, but that toIndex and fromIndex is used in HashMap. The index 0 is never used, so using fromIndex on the 0 default return value of LoadInteger results in nil.
This is only an issue with primitive types, because other types are supposed to be nil if the key does not exist.

If the cast is moved outside from the HashMap.get function to its call, the desired return type is known and nil could be changed to the correct default value. In JASS the cast is also done at the call and different functions are used to cast the integer to the correct type.

Frotty:
> I don't know much about docker, so I guess just remove the tests for now. Should I leave the functions in and just remove the test annotation or remove the function as well?

Don't remove anything, you can mark tests as ignored. What exactly doesn't work tho regarding to lua?

>Are you referring to the hashtable issue in #885 ?
The issue seems to be much more limited than I initially thought. It's not that hashtables return nil, but that toIndex and fromIndex is used in HashMap. The index 0 is never used, so using fromIndex on the 0 default return value of LoadInteger results in nil.
This is only an issue with primitive types, because other types are supposed to be nil if the key does not exist.

>If the cast is moved outside from the HashMap.get function to its call, the desired return type is known and nil could be changed to the correct default value. In JASS the cast is also done at the call and different functions are used to cast the integer to the correct type.

I don't really get the issue, so dunno.

Jampi0n:
The error is that it cannot find a lua installation:
`java.lang.RuntimeException: java.io.IOException: Cannot run program "lua": error=2, No such file or directory`
I had the same error locally until I installed lua.

The hashtable issue is more of a typecasting issue in lua and it affects some other parts as well.

As an example, take the JASS HashMap:
```function HashMap_get takes integer this, integer key returns integer```
It only deals with integers and when get is called, the arguments are converted to integer and the return type is converted to the correct type:
```indexToValueType(HashMap_get(this, keyTypeToIndex(key),...)```
The typecast functions depend on the type parameters of this HashMap instance.

In lua, the HashMap function takes and returns values of unkown type and they are cast to integer inside the function:
```
function HashMap_get(this, key) 
    return objectFromIndex(LoadInteger(this, objectToIndex(key)))
end
```

The JASS approach has the advantage, that different cast functions can be used depending on the type.
In lua this is not possible, because the casting is inside the function, so it is the same for any type.
In JASS you simply omit casting, if the type is an integer. In lua you always look up the object with objectFromIndex, which may result in nil even if the type is integer.

Currently there are other issues that are also related to typecasting in lua:

String concatenation may not be used, if one of the operands is returned by a function that returns not just strings:
```
function f()
	let strMap = new HashMap<int,string>()
	strMap.put(0, "some string")
	print("" + strMap.get(0))
```
The plus operator in the last line is not translated to string concatenation, because the return type of strMap.get is any.
In JASS, the type would be correctly identified by the compiler, because the return value of get would be wrapped inside a stringFromIndex.

Functions with type parameters are not translated correctly (the new commaList implementation does not work in lua because it uses LinkedList<T>.map<Q>).
```
public interface ConversionFunc<T, Q>
	function run(T t) returns Q

class A<T>
	T elem

	construct(T elem)
		this.elem = elem

	function convert<Q>(ConversionFunc<T, Q> itr) returns Q
		return itr.run(elem)

@compiletime function foo()
	let a = new A<int>(4)
	let b = a.convert<string>() (integer t) ->
		return I2S(t)
	DisplayTimedTextToPlayer(GetLocalPlayer(), 0, 0, 45, b)

	let c = new A<string>("9")
	let d = c.convert() (string t) ->
		return S2I(t)
	DisplayTimedTextToPlayer(GetLocalPlayer(), 0, 0, 45, I2S(0 + d))

init
	foo()
```
In this example, the closure implementations are translated correctly, so the fist closure is just `return I2S(t)` and the second one is `return S2I(t)`. However, they are wrapped by functions called `run_wrapper2` and `run_wrapper3`, resulting in:
`return stringToIndex(I2S(t)` and `return S2I(stringFromIndex(t))`. I don't know why or where these functions were put in the mapscript. In [ExprTranslation.java](https://github.com/wurstscript/WurstScript/blob/e25c260b65dcb3186cb9c6ec1e9cb44c78bb7b0d/de.peeeq.wurstscript/src/main/java/de/peeeq/wurstscript/translation/imtranslation/ExprTranslation.java#L102) they are explicitly excluded for lua.
With these functions the types are wrong, resulting in an exception at compiletime and wrong values at runtime.

These errors can easily be avoided by changing the wurst code a little bit. So I don't know if it's worth trying to fix them, considering it will probably be a lot of work and the new generics should fix the problems anyway.

Frotty:
> The error is that it cannot find a lua installation:
> java.lang.RuntimeException: java.io.IOException: Cannot run program "lua": error=2, No such file or directory
> I had the same error locally until I installed lua.

Yeah so this could be fixed with a docker image that has a java and lua installation.

Regarding the rest, I think this was one of the reasons for the new typeclasses feature and new generics, which wouldn't rely on toIndex/fromIndex hacks anymore.

Frotty:
LGTM, thanks 👍

Issue #994, opened 2021-04-10
Null references not correctly migrated from compiletime to runtime in lua


Jampi0n:
**Describe the bug**
If an object is created at compiletime and it contains a null reference,  the initCompiletimeState function will save the null reference as an integer.

Here is an example:
```
package Hello

import NoWurst
import public _Handles // map does not load without this import. error in generated lua script?
import public MagicFunctions


function print(string msg)
	DisplayTimedTextToPlayer(GetLocalPlayer(), 0., 0., 45., msg)

class A
	function foo()
		print("I am A")

class B
	A a = null

function createB() returns B
	return new B()

constant b = compiletime(createB())

init
	if b.a == null
		b.a = new A()
	b.a.foo()
```

The init part will not work, because b.a == null returns false (b.a = 0), so no new object is created and foo cannot be called on an integer.

Frotty:
the lua backend isn't finished, thanks for the report

Frotty:
is this fixed?

Jampi0n:
yes, it is.

PR #996, opened 2021-04-12
lua: use BJDebugMsg for error messages if not using ErrorHandling package

Jampi0n:
In lua, init blocks result in incorrect map scripts, if the package ErrorHandling is not used.
In particular, the print function inside the error function will be "$debugPrint". The $ character causes the map to not load.

This change is based on the jass expression translation here: [ExprTranslation.java](https://github.com/wurstscript/WurstScript/blob/e25c260b65dcb3186cb9c6ec1e9cb44c78bb7b0d/de.peeeq.wurstscript/src/main/java/de/peeeq/wurstscript/translation/imtojass/ExprTranslation.java#L28)

This change only has an effect if ErrorHandling is not used. Then the call to "$debugPrint" is replaced with "BJDebugMsg" like in the jass expression translator.

### DIFF
--- a/de.peeeq.wurstscript/src/main/java/de/peeeq/wurstscript/translation/lua/translation/ExprTranslation.java
+++ b/de.peeeq.wurstscript/src/main/java/de/peeeq/wurstscript/translation/lua/translation/ExprTranslation.java
@@ -3,6 +3,7 @@ package de.peeeq.wurstscript.translation.lua.translation;
 import de.peeeq.wurstscript.WurstOperator;
 import de.peeeq.wurstscript.jassIm.*;
 import de.peeeq.wurstscript.luaAst.*;
+import de.peeeq.wurstscript.translation.imtranslation.ImTranslator;
 import de.peeeq.wurstscript.types.TypesHelper;
 
 import java.util.Optional;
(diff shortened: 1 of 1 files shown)


Frotty:
Just for information, why wouldn't you use the `ErrorHandling` package?

Jampi0n:
I tried to construct minimal examples to reproduce errors. Using NoWurst the generated map script is smaller and easier to debug.
In this case I tried to come up with the example of #994 and noticed that the map doesn't load without `_Handles`. Then I did more tests and noticed, that it's because of `ErrorHandling`.

For normal maps I always use `ErrorHandling`.

Frotty:
How exactly does this check whether debugPrint is available?

Jampi0n:
The intermediate program already checks whether ErrorHandling is available.
[imError](https://github.com/wurstscript/WurstScript/blob/4a6902da285044b49e4c1f776893877d170937ac/de.peeeq.wurstscript/src/main/java/de/peeeq/wurstscript/translation/imtranslation/ImTranslator.java#L1636) is the function which calls the error function. It searches for the error function with `findErrorFunc`, which returns the error function from ErrorHandling if available. Otherwise a default error function will be used. This default function also contains a call to $debugPrint.
The translation from $debugPrint to BJDebugMsg was missing in lua. It exists in JASS as you can see in the link I posted above.

PR #999, opened 2021-05-28
Remove empty lines between globals in lua

Jampi0n:
The lua printer adds two line breaks after every part of the program. This makes sense for functions, but causes every second line to be empty for globals.

Before:
```
function f()
    ...
end

x = nil

y = nil

z = nil

function g()
    ...
end

A.a = nil

B.b = nil
```
After:
```
function f()
    ...
end

x = nil
y = nil
z = nil

function g()
    ...
end

A.a = nil
B.b = nil
```


Frotty:
👍

PR #1000, opened 2021-05-28
fix unused import warning with configured packages

Jampi0n:
When only configured functions (functions that are overridden by the config package) are used, Wurst will consider the configured package unused. In other cases, the function is not considered to be unused, but will be attributed to the wrong package, resulting in "the import ... can be removed, because it is already included in ...".

This can be fixed by implicitly adding the configured package to the used packages, when the config package is used.

A prime example is the ErrorHandling package. In most cases, only the configurable error function is used, so when this function is configured  most packages that import ErrorHandling will throw a warning.


Frotty:
Is this related to https://github.com/wurstscript/WurstScript/issues/990 or something different?

Frotty:
Ideally add a test for this

Issue #990, opened 2021-03-11
Wrong initialization order with configuration packages


Jampi0n:
Packages are initialized in the wrong order / an implicit cyclic dependency is not detected.

Four packages are required to reproduce:
```
package Configurable
public constant int FIRST_CONSTANT = 42
public tuple two_ints(int a, int b)
@configurable public var TWO_INTS = two_ints(0, 0)

package Something
import Configurable
public constant SECOND_CONSTANT = FIRST_CONSTANT

package Configurable_config
import Configurable
import Something
@config public var TWO_INTS = two_ints(FIRST_CONSTANT, SECOND_CONSTANT)

package Output
import Configurable
init
    print(TWO_INTS.a.toString() + ", "+ TWO_INTS.b.toString())
```

Output: 0,0

The observed initialization order is:
Something -> Configurable_config -> Configurable -> Output
SECOND_CONSTANT is initialized to 0 in Something.
In Configurable_config TWO_INTS  is set to (0,0).
Only in Configurable FIRST_CONSTANT is initialized to 42.

I assume there is an implicit dependency due to the configuration. It makes sense that all packages that import Configurable also require that Configurable_config is initialized. So Something has an explicit dependency on Configurable and an implicit one on Configurable_config. However Configurable_config also has an explicit dependency on Something, resulting in a cyclic dependency.

Note: inlining may remove the issue in this case, since the variables are inlined and never initialized

Jampi0n:
I experimented a little bit with this and I think a possible solution is to:
- add implicit imports from the configurable package to the config package

The downside of this approach is that the error message is shown for the configurable package, which would be confusing.

To fix this problem one can instead do:
- add implicit imports from the configurable package to the config package
- all packages that import the configurable package also implicitly import the config package

With that, the error message will be shown for the config package.

I have a working implementation of that approach, but the error message in vscode is not always updated properly. That seems to also be the case with regular imports though, so maybe it's fine. Should I make a PR for this?

PR #1001, opened 2021-05-28
detect cyclic dependencies caused by config packages 

Jampi0n:
Fix for #990 .

There is an implicit initialization order dependency between the config and configured package:
The config package is always initialized directly before the configured package. However, this dependency is not known to the wurst validator, so it is possible to create cyclic dependencies and packages will be initialized before their imports.

This PR implicitly merges the dependencies on the config package.

- cyclic dependency error message is displayed for the config package
- the config package can still import the configured package. This is necessary in order to use definitions from the configured package

This can break existing and working maps. So implementing the stricter cyclic dependency check as a warning may be preferred. At least for some time, similar to globals being used above their declaration.

### DIFF
--- a/de.peeeq.wurstscript/src/main/java/de/peeeq/wurstscript/attributes/InitOrder.java
+++ b/de.peeeq.wurstscript/src/main/java/de/peeeq/wurstscript/attributes/InitOrder.java
@@ -7,13 +7,12 @@ import com.google.common.collect.Sets;
 import de.peeeq.wurstscript.ast.WImport;
 import de.peeeq.wurstscript.ast.WImports;
 import de.peeeq.wurstscript.ast.WPackage;
-import de.peeeq.wurstscript.utils.Utils;
 
 import java.util.Collection;
 import java.util.List;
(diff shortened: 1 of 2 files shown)


Frotty:
Nice 👍 

> So implementing the stricter cyclic dependency check as a warning may be preferred

This sounds reasonable.
Probably time to turn the global warning into an error now.

Jampi0n:
Cyclic dependencies caused by config dependencies are now only warnings. The check runs now twice, once the old version with errors and once the new version with warnings. Reported errors are stored, so they are not reported again as a warning.
I added some more functions in order to avoid code duplication.

The warning message will show that the configured and config package belong together. That should make it clear that imports from the configured package are also considered for the cyclic dependency check.
`Test/Test_config -> Requirement -> Test/Test_config`

Frotty:
Thanks 👍 
We should probably change the "initlater" phrase to also include proper code organization - many ppl get some issues after heavy initlater usage, since they don't understand what they are doing.

PR #1004, opened 2021-06-20
optimizers for lua

Jampi0n:
lua translation does not eliminate as many constructs from the intermediate language as JASS. As a result, the optimizers need to support these constructs:
classes, methods, vararg loops


Frotty:
Thanks, worked in my tests.
The static variable for class type is not ideal, but I'm happy to merge working stuff.

PR #1025, opened 2021-08-16
Optimize compilation to Jass

peq:
some minor optimizations:
- updated AST library, which improves removeAll calls
- optimize replaceBy method to smarter search for position of element in
parent


Frotty:
LGTM, what's missing?

peq:
There are some differences in the generated code. Seems the optimization behaves differently now, which I did not expect. Need to check which version is right or if the differences are just normal nondeterminism problems.

PR #1043, opened 2021-12-12
Modularize application (use Java module system)

peterzeller:
With the modularization, we can use the jlink command (`gradle jlink -x test`) to build a package including a custom Java Runtime Environment.

This can later be used to automatically download the correct version of Java with the application directly from the VSCode plugin.

Pros:

- Easy install (just install the plugin)
- No more problems with wrong Java installations
- We can update Java without breaking it for users

TODO:

- [ ] Importing the setup tool did not work because of conflicts, so I included the necessary files as decompiled Java files. It's probably better to move these files to a common shared library.
- [ ] Test it.


Frotty:
The part of building a minimal jre via jlink is now implemented.
Modularization is not needed in our case, because we don't need to be truly modular or hide any packages - we can just use jdeps to find the needed modules and then use jlink to build the minimal jres for each platform.
The gradle scripts & github workflows do exactly that.
In the next vscode extension release, the extension will download the runtime and compiler from GitHub releases and use that instead.

Issue #998, opened 2021-05-08
Function calls as array index are executed twice when assignment shorthands are used


Jampi0n:
A function call that is used as array index for assignment shorthands (+=, -=, ++, ...) is executed twice:
`test[getIndex()]++` gets translated to `test[getIndex()] = test[getIndex()] + 1`
The function can return different values in each call and could also have side effects that are repeated.

The expected behavior would be to have a temporary variable which stores the return value and use that variable as index:
```
tmp = getIndex()
test[tmp] = test[tmp] + 1
```

PR #1061, opened 2022-07-02
Temporary index variable for shorthand assignments

Jampi0n:
Addresses #998 .

rewrites
`test[getIndex()]++`
to
```
let tmp = getIndex()
test[tmp] = test[tmp] + 1
``` 

Few remarks:
- I rewrote transformStatement so it returns a list of statements rather than a single one in order to add the temporary variables before the set statement. Maybe there is a way to combine multiple statements in a single one?
- It assumes, indexes are integers. Since this transformation happens so early in the compiler, there doesn't seem to be a way to get the type of the index expressions. Should be fine though, because all array indexes are always integer.
- From a quick look, jurst seems to have the same code. I haven't updated it, but if this implementation is ok, I can apply it to jurst as well.
- I used tmpArrayIndex_ combined with the line number as name for the temporary variable which should avoid conflicts, but if there is a function to get a unique identifier that would be better.

Integer literals and variables don't use temporary variables:
`test[4]++` -> `test[4] = test[4] + 1`
`test[myInt]++` -> `test[myInt] = test[myInt] + 1`
When using expression that can be evaluated to constants, the optimizer will take care of that and you end up with a case like above.
Right now integer arrays are also stored in temporary variables:
`test[myIntArray[0]]++` ->
```
let tmp = myIntArray[0]
test[tmp] = test[tmp] + 1
```
In theory one could leave integer arrays, as long as the index itself is again constant, but if enough integer arrays are used, using a temporary variable will be faster. Could probably add a cutoff and say if it's less than x nested integer arrays, keep them, otherwise use temp variable.

### DIFF
--- a/de.peeeq.wurstscript/src/main/java/de/peeeq/wurstscript/parser/antlr/AntlrWurstParseTreeTransformer.java
+++ b/de.peeeq.wurstscript/src/main/java/de/peeeq/wurstscript/parser/antlr/AntlrWurstParseTreeTransformer.java
@@ -19,7 +19,6 @@ import org.antlr.v4.runtime.tree.ParseTree;
 import org.antlr.v4.runtime.tree.TerminalNode;
 import org.eclipse.jdt.annotation.Nullable;
 
-import java.util.LinkedList;
 import java.util.List;
 
 public class AntlrWurstParseTreeTransformer {
@@ -478,7 +477,7 @@ public class AntlrWurstParseTreeTransformer {
     private WStatements transformStatementList(List<StatementContext> stmts) {
         WStatements result = Ast.WStatements();
         for (StatementContext s : stmts) {
-            result.addAll(transformStatement(s));
+            transformStatement(result, s);
         }
         return result;
     }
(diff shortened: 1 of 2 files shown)


Frotty:
Thanks for the PR.

> I rewrote transformStatement so it returns a list of statements rather than a single one in order to add the temporary variables before the set statement. Maybe there is a way to combine multiple statements in a single one?

Could be fine but I think the current implementation might could be improved, as it currently would mean a list allocation for every statement, regardless whether it needs more than 1 return statement.

> I used tmpArrayIndex_ combined with the line number as name for the temporary variable which should avoid conflicts, but if there is a function to get a unique identifier that would be better.

Yea might be better to use `NameProvider` or something.

> In theory one could leave integer arrays, as long as the index itself is again constant, but if enough integer arrays are used, using a temporary variable will be faster. Could probably add a cutoff and say if it's less than x nested integer arrays, keep them, otherwise use temp variable.

Yea could be a later optimization but doesn't have to be in this PR.

Frotty:
Very nice, thank you 👍

PR #1062, opened 2022-07-03
Use correct timeTaker for compiler

Jampi0n:
Run times of local optimizations were missing.


null15:
> Run times of local optimizations were missing.

Good job Jampy

Frotty:
Thanks for the fix 👍

PR #1063, opened 2022-07-04
Improve Lua optimizers

Jampi0n:
Bugfixes and improvements for Lua optimizers.
Implements some handling of tuples and members that was missing and could cause:

<details> 
  <summary>Compiler error due to unattached variables, when doing some optimizations with tuples</summary>
-inline
-localOptimizations
-lua

```
package MyTest

function locals() returns bool
    var d = vec2(1,0)
    d.y = 2
    return d.x == 1
    
init
    if locals()
        print("Success")
```
</details>

<details> 
  <summary>Wrong output script due to incorrect assumptions by the optimizer </summary>
-inline
-localOptimizations
-lua
   
```
package ListSort

import LinkedList

init
    let list = new LinkedList<int>()
    list.add(1,2)
    list.sort()
    print(list.get(0))
    print(list.get(1))
```

This prints 1 and 0. It is caused by the split function in LinkedList:
```
...
let tmp = halfRef.next
halfRef.next = dummy
return tmp
```
Without stacktraces, only the member set halfRef.next is between the tmp accesses and the optimizer would think tmp equals halfRef.next, so it can just return halfRef.next.
</details>

The optimizer now also deals better with tuple literals and avoids copying them after constructing them. There are still possible improvements for tuples, but I first wanted to fix the bugs.

Unused members are now removed like in Jass.

Some Lua translation steps that relied on type information were moved before local optimizations:
- null string to ""
- string concatenation

This makes it possible to unify some types before the optimizers, allowing more local merging.
Arrays and tuple types are kept, because they are required for the initialization.

I actually only found the first issue by running all unit tests with Lua and getting the error. With these changes all unit tests also passed in Lua with the same optimizer options as in Jass, as long as execution is disabled for Lua. Maybe it would be good to have unit tests by default also compile with Lua just so compiler errors would be detected.

### DIFF
--- a/de.peeeq.wurstscript/src/main/java/de/peeeq/wurstio/WurstCompilerJassImpl.java
+++ b/de.peeeq.wurstscript/src/main/java/de/peeeq/wurstio/WurstCompilerJassImpl.java
@@ -857,6 +857,11 @@ public class WurstCompilerJassImpl implements WurstCompiler {
             printDebugImProg("./test-output/lua/im " + stage++ + "_afterinline.im");
         }
 
+        // eliminate local types
+        beginPhase(6, "eliminate local type");
+        getImProg().flatten(imTranslator2);
+        EliminateLocalTypes.eliminateLocalTypesProg(getImProg(), imTranslator2);
+
         optimizer.removeGarbage();
         imProg.flatten(imTranslator);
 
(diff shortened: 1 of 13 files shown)


Frotty:
Cool that you still finished it, massive props 👍 I plan to use it for jass to lua map optimization in w3p.

> Maybe it would be good to have unit tests by default also compile with Lua just so compiler errors would be detected.

Yea, could be enabled by default. I guess only the lua execution binary is the problem?

Jampi0n:
> I guess only the lua execution binary is the problem?

Executing Lua with standard library generally doesn't work, because it uses natives (e.g. InitHashtable for compiletime state).
With the above two changes and disabling Lua execution while standard library is enabled, all but one test pass.
The test that fails is enum_to_int_test also fails, because enum to int cast works differently in Lua and the third value in the enum is not necessarily 2. That's probably something that should be changed as well, considering the manual states:
> The coalescent integer value starts at 0, incrementing with each succeeding enum member. So for MyUnitState FLYING will be 0, GROUND 1 and WATER 2.

PR #1064, opened 2022-07-12
Remove type dependency from cast translation

Jampi0n:
What I said [here](https://github.com/wurstscript/WurstScript/pull/1063#issuecomment-1179729151) with the enums working differently is wrong. I assumed so, because I saw objectToIndex being used for the cast.
What happened is that due to the local type elimination, the translator did not know it was a cast from int (Enum) to int and only saw a cast from something to int, so it inserted the objectToIndex.

Now these unnecessary casts from int to int are removed in advance, so that the translator does not need to know the type of the expression that's being cast and works with eliminated local types.


Frotty:
Can we add a test for this?

Jampi0n:
I added two unit tests. The first checks if string concatenation is correctly translated.
The second one checks if type casting is correctly translated for both enums (int to int) and classes (with toIndex and fromIndex).
I used local variables to check if it still works after local types are eliminated.

PR #1095, opened 2025-09-17
Make pretty printer fail on compilation errors

Cokemonkey11:
- Modified PrettyUtils to check for compilation errors before formatting
- Added validation step that runs compiler.checkProg() after parsing
- Throws RuntimeException with error details if compilation errors found
- Added comprehensive test coverage for both valid and invalid code
- Ensures pretty printer fails fast with clear error messages


Cokemonkey11:
As discussed, there's a lot that's directionally wrong about this

PR #1203, opened 2026-07-20
Add SQLite compiletime natives

Donach:
## Summary

Add SQLite-backed compiletime natives to the interpreter:

- open and close database connections
- prepare, bind, step, reset, and finalize statements
- read integer, real, and string result columns
- execute direct SQL statements

The API uses integer handles at the Wurst boundary while keeping JDBC resources encapsulated in `CompiletimeNatives`. Bind indexes follow JDBC's 1-based convention; result column indexes are 0-based for Wurst array-style access.

## Motivation

Wurst projects with larger structured content catalogs currently have to generate or parse intermediate text before compilation. Compiletime SQLite access lets them query a reviewed SQLite catalog directly while object generation and compiletime registration run.

The compiler already ships `sqlite-jdbc`; this change exposes that existing capability to compiletime Wurst code without adding a dependency.

## Resource and error handling

- statements are tracked against their owning connection
- closing a connection finalizes its remaining statements
- reset closes the active result set while retaining the prepared statement
- invalid handles produce explicit interpreter errors
- JDBC cleanup failures are reported rather than silently ignored
- direct execution uses try-with-resources

## Tests

The interpreter-level test exercises an in-memory database end to end, including:

- schema creation and direct execution
- prepared statements
- integer and string binding
- stepping and column access
- reset, finalization, and connection close

Validation run against upstream `50080cdb9f4e8dda9361feb40a42f7bdfb8f6f21`:

- `./gradlew test --tests tests.wurstscript.tests.CompiletimeTests.testCompiletimeSQLite`
- `./gradlew test --tests tests.wurstscript.tests.CompiletimeTests --tests tests.wurstscript.tests.CompiletimeNativesTest`
- `./gradlew shadowJar --no-daemon`
- `./gradlew test --no-daemon` — full suite passed


Donach:
Closing in favor of #1180. The clean implementation, handle validation, error handling improvements, and connection/statement lifetime management from this PR have been consolidated into #1180.

PR #1208, opened 2026-07-31
Fix SCC dataflow worklist ordering

Frotty:
## Summary

- process statements within each strongly connected component in forward discovery order so dataflow facts propagate with the CFG
- deduplicate queued statements and use constant-time SCC membership checks
- report a source-located internal compiler error and skip only the affected analysis body if the convergence watchdog is ever reached
- add a regression with sixteen mutated locals inside a loop

## Root cause

The Gabow SCC implementation emits components in reverse topological order and emits the nodes inside each component in reverse discovery order. The executor corrected the component order but retained the reversed node order. For the reported 17-node loop SCC, valid dataflow analysis required 561 worklist iterations while the watchdog allowed only 389, producing a false non-convergence failure. Forward node order reduced the same analysis to 51 iterations.

## Impact

Large loop bodies no longer incur near-quadratic backward propagation or fail the convergence heuristic. Any future watchdog failure is attached to the analyzed source body and does not abort validation of unrelated functions.

## Validation

- `./gradlew test --tests "tests.wurstscript.tests.FlowAnalysisTests.dataflowConvergesForManyMutatedLoopLocals"`
- `./gradlew test --tests "tests.wurstscript.tests.FlowAnalysisTests"`
- `./gradlew test` — 1,455 tests, 0 failures, 0 errors


Frotty:
@codex review

PR #1210, opened 2026-07-31
Enforce Jass vararg parameter limit

### LINKED ISSUE #729: vararg is limited to be size of 31
as the wurst manual says that "vararg" is unlimited in amount of values inside, see:
https://wurstlang.org/manual.html#extension-functions
"We can pass any amount of integers to the function"

it should be noted that the generated jass code can only handle up to 31 parameters, and therefore you cannot used more than 31 values for the vararg.

Frotty:
Fixes #729.

Vararg calls now fail at the source location when lowering would exceed Jass's 31-parameter limit. Receivers and constructed objects count toward the generated total, and the diagnostic suggests multiple/cascade calls or passing a collection.

Tests cover free functions, methods, constructors, and both sides of the boundary. The full Gradle suite passes.


Frotty:
@codex review

PR #1211, opened 2026-07-31
Add readonly field modifier

### LINKED ISSUE #217: Add "publicread" keyword for class members.
Add a feature that allows public reading of classmembers but only protected/private write access to avoid excessive use of gettermethods which blow up the code.

Solution 1: introduce a new keyword for that purpose:

```
class MyClass
    // missilerange can only be written from inside the class, but read from everywhere:
    publicread real missileRange
```

Solution 2: make writing gettermethods easier by adding some kind of shortcut (exemplary syntax):

```
class MyClass
    private real missileRange : public getMissileRange
```

compiles into:

```
class MyClass
    private real missileRange
    public function getMissileRange() returns real
        return missileRange
```

Frotty:
Replaces the unused `publicread` keyword with a composable `readonly` modifier. Read visibility follows the existing visibility modifier, while writes are restricted to the declaring class, module, or package.

Closes #217.

Tests: `./gradlew test` (1,478 passed)


Frotty:
@codex review

PR #1215, opened 2026-08-01
Propagate Lua target to compiletime interpreter

Frotty:
## What changed

- Initialize `MagicFunctions_isLua` in the compiletime interpreter from the translator target.
- Use an untracked interpreter write so the backend constant is not treated as user-mutated scalar state or replayed at runtime.
- Evaluate lazy global initializers in an untracked scope so their ordinary side effects are not persisted and duplicated when runtime initialization repeats them.
- Add Jass/Lua regressions for backend detection and lazy-initializer side effects.

## Supported migration boundary

Compiletime-to-runtime state migration covers writes performed explicitly by compiletime functions.

Side effects encountered only while lazily evaluating a global initializer are intentionally not recorded for migration, because runtime executes that initializer again. Compiletime-only side effects hidden inside a lazy initializer are therefore unsupported; persistent mutation should be performed by an explicit compiletime function after initialization.

This boundary avoids speculative runtime re-execution and keeps initializer ordering and side-effect behavior deterministic.

## Why

`-lua` correctly reached `RunArgs` and `ImTranslator`, but the compiletime interpreter lazily evaluated the source declaration of `MagicFunctions.isLua`, whose Jass-safe default is `false`. Compiletime functions therefore observed Jass mode even for Lua builds.

Separately, tracking side effects produced by lazy initializer evaluation caused those writes to be replayed before runtime repeated the initializer, duplicating state.

## Validation

- `LuaBackendAuditTests.compiletimeInterpreterSeesLuaTarget`
- complete `CompiletimeTests`
- complete `LuaBackendAuditTests`
- complete Gradle test suite
- `git diff --check`


Frotty:
@codex review

Frotty:
@codex review

Frotty:
@codex review

Frotty:
@codex review

Frotty:
@codex review

Frotty:
@codex review

Frotty:
@codex review

Frotty:
@codex review

Frotty:
@codex review

Frotty:
@codex review

Frotty:
@codex review

PR #1216, opened 2026-08-02
Fix generic field toIndex assignments

### LINKED ISSUE #647: Cannot assign to generic fields with toIndex function
```
    @Test
    public void genericToIndexBug2() {
        testAssertOkLines(true,
                "package test",
                "native testSuccess()",
                "class C<T>",
                "    T x",
                "function stringToIndex(string s) returns int",
                "    return 42",
                "function stringFromIndex(int i) returns string",
                "    return \"42\"",
                "init",
                "    C<string> c = new C<string>",
                "    c.x = \"42\"",                           // translation fails here
                "    if c.x == \"42\"",
                "        testSuccess()");

    }
```

Frotty:
## Summary

- preserve the raw expected type when translating assignment right-hand sides
- restore legacy generic `toIndex` conversion for field assignments
- cover the exact issue reproduction on Jass and Lua

## Root cause

Assignment expected-type inference normalized the target type before IM translation. For a legacy generic field, that erased the bound type parameter which carries the `stringToIndex` conversion, so Jass emitted a string assignment into an integer-backed field.

## Validation

- `./gradlew test --tests "tests.wurstscript.tests.GenericsTests.genericToIndexFieldAssignment" --tests "tests.wurstscript.tests.LuaTranslationTests.legacyGenericToIndexFieldAssignmentRoundTripsInLua"`
- `./gradlew test --tests "tests.wurstscript.tests.GenericsTests" --tests "tests.wurstscript.tests.LuaTranslationTests"`

Closes #647

### DIFF
--- a/de.peeeq.wurstscript/src/main/java/de/peeeq/wurstscript/attributes/AttrExprExpectedType.java
+++ b/de.peeeq.wurstscript/src/main/java/de/peeeq/wurstscript/attributes/AttrExprExpectedType.java
@@ -37,7 +37,7 @@ public class AttrExprExpectedType {
             } else if (parent instanceof StmtSet) {
                 StmtSet stmtSet = (StmtSet) parent;
                 if (stmtSet.getRight() == expr) {
-                    return stmtSet.getUpdatedExpr().attrTyp();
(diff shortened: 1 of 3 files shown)


Frotty:
@codex review

Issue #940, opened 2020-02-14
String Interpolation


peq:
Add a nicer way to construct strings.

So instead of 

    return "(" + v.x.toString() + ", " + v.y.toString() + ", " + v.z.toString() + ")"

we could write:

    return s"(${v.x}, ${v.y}, ${v.z})"

I am not sure if we should add a special syntax for strings with interpolation (like the `s` prefix in the example above) or just allow interpolation in every string (and use `\$` to write a real dollar sign).

We should also add a syntax for multiline strings, like:

    let s = """This is a
                   string with
                   several lines"""

This thread is open for suggestions and discussion.

Jaccouille:
Hello, I'd like to contribute to the multiline strings, may I ask for some pointers?

What I believe I've understood so far I don't have any experience with lexer/parser :
- [Wurst.g4](https://github.com/wurstscript/WurstScript/blob/e6f36a8fdef6e01ab005bc27e08f6ed68e055cc5/de.peeeq.wurstscript/src/main/antlr/de/peeeq/wurstscript/antlr/Wurst.g4) contains antlr rules responsible for parsing
- [AntlrWurstParseTreeTransformer](https://github.com/wurstscript/WurstScript/blob/e6f36a8fdef6e01ab005bc27e08f6ed68e055cc5/de.peeeq.wurstscript/src/main/java/de/peeeq/wurstscript/parser/antlr/AntlrWurstParseTreeTransformer.java#L24) replace the tokens with an output

I don't know how to test apart from running `make_for_userdir`, reload and build in Visual studio code.
I don't know how to test the language server so I can see that my multi line string is recognized as a string in Visual studio code.

Frotty:
Instead of building the compiler, I would suggest you write unit tests instead, which you can run in your IDE. Then you can go from there to see what errors you have to address. Tests will be required anyway for the PR to be accepted.
Otherwise, you seem to be on the right path, you need to add an antlr rule (perhaps backticks would be better than """).
Then probably the easiest way would be to transform the special multiline string into a concatenation of regular strings with line breaks added.
Regarding Visual Studio Code, the syntax highlighting is defined in the extension's source repository.

PR #1217, opened 2026-08-03
Infer toString in string concatenation

Frotty:
## Summary

- infer visible zero-argument `.toString()` conversions in string-concatenation chains
- preserve applicable `op_plus` overload precedence and use implicit conversion only as fallback
- preserve normal member-method precedence over extension methods before validating the conversion return type
- support class methods, extension methods, generic classes, and both Jass and Lua backends
- warn when an explicit `.toString()` is truly redundant and provide a quick fix that removes only the call suffix
- suppress this migration warning for imported libraries so existing dependencies do not produce warning spam
- track inferred extension imports, reject uninferable generic conversions, and diagnose ambiguous conversions
- preserve imported-library classification when dependency units are reparsed by the language server
- pin wc3libs by full commit SHA so cold-cache JitPack resolution works in Ubuntu CI

## Example

```wurst
"player: " + player
```

now resolves the same `player.toString()` conversion that an explicit call would use, unless an applicable operator overload takes precedence.

## Validation

- focused failing regressions for every review finding
- Jass and Lua behavioral coverage for interpolation and operator precedence
- complete expression, operator-overloading, model-manager, and language-server test classes
- `packageSlimCompilerDist --refresh-dependencies`
- complete Gradle test suite

Addresses #940

### DIFF
--- a/de.peeeq.wurstscript/build.gradle
+++ b/de.peeeq.wurstscript/build.gradle
@@ -109,7 +109,7 @@ dependencies {
     implementation 'com.github.albfernandez:juniversalchardet:2.4.0'
     implementation 'org.xerial:sqlite-jdbc:3.46.1.3'
     implementation 'com.github.inwc3:jmpq3:e28f6999c0'
-    implementation 'com.github.inwc3:wc3libs:ac41f780a5'
+    implementation 'com.github.inwc3:wc3libs:ac41f780a5e2dfc35310be4ed3267f23ab3fea44'
     implementation 'com.github.wurstscript:wurst-project-config:348fcd4ef5'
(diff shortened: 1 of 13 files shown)


Frotty:
@codex review

Frotty:
@codex review

Frotty:
@codex review

Frotty:
@codex review

Frotty:
@codex review

Frotty:
@codex review

Frotty:
@codex review

Frotty:
@codex review

Frotty:
@codex review

PR #1220, opened 2026-08-08
Ship compiler language digest for agents

Frotty:
## Summary

- add `agent-docs/WURST_LANGUAGE.md` to compiler resources
- preserve the compact language syntax and semantic guidance removed from the generated project preset
- keep dependency, UI, and object-editor policy out of the compiler-owned language reference

## Why

WurstSetup PR #77 makes generated `AGENTS.md` compact and points to a compiler-matched local language digest. This resource is the missing source for that path; without it, language guidance would disappear behind an unavailable file and web fallback.

Grill extracts this resource to `~/.wurst/wurst-compiler/agent-docs/WURST_LANGUAGE.md` when the compiler is installed or verified.

## Validation

- `./gradlew processResources`
- verified `build/resources/main/agent-docs/WURST_LANGUAGE.md` is produced
- `git diff --check`


Frotty:
@codex review

PR #1229, opened 2026-08-15
Make type substitution carry what is known about a type

Frotty:
Substitution of type variables was expressed as two parallel lists — the type variables of a
function or class next to the type arguments given at a use of it — paired up positionally at each
point of use:

```java
int index = typeVars.indexOf(t.getTypeVariable());
...
return generics.get(index).getType();   // unwraps, drops everything else
```

An `ImTypeArgument` is a type *plus* what is known about it, but pairing the lists unwrapped the
argument and returned the bare `ImType`. The type class binding lives on the argument, so it could
not survive substitution. Callers that needed it re-attached it by hand; the ones with nowhere to
re-attach it from silently produced an argument with no instance.

This replaces that with one primitive, `TypeSubst`. Substituting into an argument position now
yields the argument that was bound, so the instance travels with the type it belongs to. Lookup is
by identity, which is what positional lookup already did — IM nodes do not override `equals`.

### What it fixes

A type class bound is now usable from inside a closure:

```wurst
interface Producer
    function produce() returns int

function indexLater<T: Indexable>(T x) returns Producer
    return () -> T.toIndex(x)
```

Two things were wrong. Substitution dropped the binding on the way into the closure class, and
`TypeRewriter` never rewrote the one type variable reference that is held directly rather than as a
type — the one on `ImTypeVarDispatch`. A closure lifting its body into a class of its own therefore
left the dispatch naming a variable of the function it had left, with nothing able to bind it.

Jass only for now. Lua keeps generics erased and specialises what it can reach through a concrete
type; a closure is reached through its interface, so the specialised class gets built but nothing
calls it. That is a change to Lua's erasure rather than to substitution, and is left alone here —
`dispatchInsideClosureIsRejectedForLua` pins that it is reported clearly rather than mistranslated,
and becomes the success case if that is ever fixed.

### Also

- Drops `ProgramState.substituteTypeVars`, a second implementation of substitution that disagreed
  with the first about whether bindings survive, and was only ever reached from itself. It also
  silently did not recurse into array and tuple types.
- `substituteType` no longer rebuilds a type when there is nothing to replace. It runs for the
  return type of every call, and was allocating an equal copy of every class and tuple type.

### Verified

- Full suite: **1646 tests, 0 failures** (1634 before, plus 10 unit tests for the primitive and 2
  for the closure case).
- Emitted Lua byte-identical to `0d2bb0fbd` across the tests that produce it (28 files, no
  differences), and byte-identical across two runs of this branch — AGENTS.md §8.
- `TypeSubstTests` covers the primitive directly: binding survives in argument position, through
  nesting, and applied to an argument; a binding already present wins; variables sharing a name stay
  distinct; a missing argument is tolerated until it is named.

Worth knowing, unrelated to this change: emitted **Jass** is not stable across test runs. Comparing
two runs of `0d2bb0fbd` against each other gives 91 differing `.j` files, the same magnitude as this
branch against itself. The differences are temp names (`temp151` vs `temp8`) from a counter that is
not reset between compilations in one JVM, so the output depends on how many tests ran before. It is
a test-harness artefact rather than a defect in compiling a single map, but it does mean `.j` output
cannot be diffed across runs to check a change.

### Not in scope

The named follow-ups from #1226 stand: Lua constructor and closure dispatch, module bounds, and the
stdlib conversion of `HashMap`/`HashList`/`HashSet`/`Table` to use `Indexable`. One name-tolerant
type-variable lookup remains in `EliminateGenerics.inheritTypeClassBinding`, covering the in-place
rewrite of call-site type arguments; it can go once a type parameter is guaranteed one `ImTypeVar`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,18 @@
     An instance of `I` for type `X` may only be declared in the package declaring `I` or the one declaring
     `X`, and only once, so `I` for `X` means the same thing throughout a program regardless of imports.
 
+- A type class bound is now usable from inside a closure, so a bounded generic can hand work to one:
+
+        interface Producer
+            function produce() returns int
+
+        function indexLater<T: Indexable>(T x) returns Producer
+            return () -> T.toIndex(x)
+
+    Substituting a type variable now carries the instance chosen for it along with the type, rather than the
+    type alone, so lifting a body into a class of its own no longer loses it. Jass only for now: Lua reaches
+    such a class through its interface and still reports the bound as unresolvable there.
+
 - Added new pseudo-natives for debugging memory leaks:
 
         // returns the maximum type id, can be usd to
--- a/de.peeeq.wurstscript/src/main/java/de/peeeq/wurstscript/intermediatelang/interpreter/ProgramState.java
+++ b/de.peeeq.wurstscript/src/main/java/de/peeeq/wurstscript/intermediatelang/interpreter/ProgramState.java
@@ -461,59 +461,6 @@ public class ProgramState extends State implements AutoCloseable {
     }
 
 
-    // Helper method to substitute type variables
-    private ImType substituteTypeVars(ImType type, Map<ImTypeVar, ImType> substitutions) {
-        return type.match(new ImType.Matcher<ImType>() {
-            @Override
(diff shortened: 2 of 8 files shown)


Frotty:
@codex review

PR #1230, opened 2026-08-15
Sanitise Lua method names where they are assigned

Frotty:
A method name becomes a Lua table key, so it has to be an identifier. A method specialised with two type arguments was named after them, commas included, and emitted

    Class.get_specialized_integer, integer = impl

which is valid Lua assigning to two targets — so the syntax check passed and the wrong thing happened quietly. A tuple argument produced characters `luac` rejects outright.

## What changed

`LuaDispatchPreparation.normalizeMethodNames` is the pass that gives one name to a whole dispatch group, so it sanitises **before** uniquing: two names that differed only in characters Lua has no place for still get a slot each. The backend maps every slot key and every `LuaMethod` name through the same function, so call sites and class tables keep agreeing. Lua's identifier rule now has one home, `LuaIdentifiers`.

## Why the assertion is part of this

The luac check never caught this, because the broken output parses. So the check is on the names themselves: every emitted function, method, variable, field and call-by-name must match `[A-Za-z_][A-Za-z0-9_]*`, verified for every `testLua` compile. That is what stops this class of bug returning rather than the fix itself.

Repro that failed before and passes after: `FastHashMapTests.tupleKeyLua`.

Full test suite green.


Frotty:
Addressed: keywords are now rejected by `isValid` and mapped out of the way by `toIdentifier`.

Confirmed the path first — it is reachable. Method names were already safe, because `normalizeMethodNames` uniques against the reserved set. A closure is not: `addDirectAliases` adds the name it implements as a dispatch alias directly, so an interface method declared `repeat` arrives as a bare keyword and is emitted as a table key. Repro fails on the previous commit with `luac: <name> expected near 'repeat'` and passes now, kept as `TypeClassTests.closureImplementingALuaKeywordName`.

The sharper half of your point was the second one: the assertion added in this PR accepted the name, and catching this before the syntax check is the whole reason it exists. It rejects keywords now.

Mapping is by trailing underscore rather than a counter, so a keyword maps to the same name wherever it is derived — call sites and class tables have to agree without consulting each other.

@codex review

Frotty:
Removed, and you were right that it should never have been there. Both `.claude/ralph-loop.local.md` and `LOOP.md` are gone from all six branches, with `/.claude/` and `/LOOP.md` added to `.gitignore` so neither comes back when the tooling rewrites them.

For the record on how they got in: neither file was touched by any commit in this work. They arrived with `e4da96d7d`, an existing commit on the branch this stack was cut from, and I did not check what that base was carrying before opening the PRs. Neither is in master, so removing them nets to nothing in the diffs — all six now contain only compiler and test files.

@codex review

PR #1231, opened 2026-08-15
Bound the wait on the Lua interpreter in tests

Frotty:
Stacked on #1230.

The Lua execution path read the spawned interpreter's stderr to EOF before touching stdout, and never bounded the wait. A program that fills the stdout pipe blocks writing while the harness blocks reading stderr, and the suite stops with no output, no timeout and no failing test. This was found the hard way: a stray worker JVM was still sitting on the build directory twenty minutes after the run appeared to hang, and a thread dump pointed at the read.

`checkLuaSyntax`, ten lines further down the same file, already drained both pipes on their own threads and waited with a timeout. The execution path now uses the same helper, so a program that does not terminate fails its own test instead of the run.

Verified against a program that reproduced the hang: indefinite before, fails in 8s after.

Full test suite green.


Frotty:
Both addressed.

**P1 — regression tests.** `LuaRunnerTests` covers each failure mode: a program that loops forever now fails its own test instead of the run, and one that prints twenty thousand lines still finishes. The second is the actual deadlock — read one stream after the other and the program blocks writing stdout while the runner blocks reading stderr. Both are Lua only on purpose: run through the Jass configurations, the non-terminating one would hang the interpreter instead, which is the same problem somewhere this change does not reach. The timeout is overridable so they take seconds rather than a minute.

**P2 — capped retention.** Done, and the P1 test immediately caught that capping alone is wrong: `testSuccess` is printed *last*, well past 64KB, so the cap discarded the one line the runner scans for and every executing test would have started failing. The watched line is now recognised while draining instead of read back out of the retained text — better regardless, since whether a program succeeded no longer depends on how much of its output was kept.

Draining itself is unchanged and still unbounded, because stopping is what blocks the process.

@codex review

Frotty:
Updated from master (#1230 is in) and the new finding is addressed.

You were right on both halves, and the second half is the one that mattered: checking the limit once per line bounds nothing when a program prints one enormous line, and `readLine()` materialises the whole line before any check here could apply — so the limit could not have helped even in the case it did cover.

Draining is by fixed-size chunks now. Retained output stops at the limit part way through a chunk, and the watched line is recognised as characters arrive, keeping no more of a line than the length of what is being looked for: past that length the line cannot be it, so none of it is worth holding. Nothing in the collector now holds more than the retained limit plus one chunk.

`LuaRunnerTests` has a third case for it — one line of about a quarter of a million characters, which the previous revision would have retained in full. All three pass, and the executing suites still detect success normally.

@codex review

Frotty:
Took one, declined the other.

**P3 (truncation notice) — taken.** Real and cheap: truncation was only noticed on a *later* chunk, so a stream whose final read crosses the limit ended at exactly 64KiB with nothing saying so — output that reads like everything the program had to say. Now noticed within the crossing chunk.

**P2 (make the huge-line test observe the bound) — declined, and the test relabelled instead.** You are right that it does not observe retention and would pass against the old collector; that was a fair hit on the test rather than the code. But asserting it needs test-only plumbing to expose the retained buffer, for a program printing a megabyte with no newline — which is not a shape this suite is going to see. Rather than leave a test claiming coverage it does not have, it is now `aProgramPrintingWithoutNewlinesStillFinishes`, documented as checking that it finishes and explicitly not the bound.

The bounded drain stays either way, since it costs nothing and the two tests above it cover the failure that actually bit us — the deadlock.

@codex review

PR #1232, opened 2026-08-15
Let declared names decide dispatch slots, and finish the FastHashMap proof

Frotty:
Stacked on #1231.

## Dispatch slots

A method claims the slots of every method it might be an override of, and `sharesSemanticName` accepted a match on **either** the declared name or the segment after the last underscore of the mangled one. For a specialised method that segment is a fragment of the type argument: `get` and `slotFor` of `FastHashMap<int, int>` both read as `integer`, and since specialisation gave them the same signature too, `get` claimed `slotFor`'s slot. The class table bound `FastHashMap_slotFor_specialized_integer__integer` to `FastHashMap_get_specialized`.

Declared names now settle it whenever both methods have one. The segment stays as the fallback for closures and bridges, which have no declaration to ask — that is the case it was there for.

The emitted Lua is the only record of which implementation a slot should hold, so the test asserts it there: every slot named after a method must bind an implementation named after that same method. It fails on the old code with exactly the binding above.

## FastHashMap

`remove` leaves a tombstone rather than an empty slot, because a probe that stopped at one would miss keys put down beyond it. `slotFor` passes tombstones over when searching and returns the first one when putting, so removing and re-putting a key reuses its slot. The probe is bounded by capacity, so a table full of tombstones cannot spin.

The cost claim is asserted on the **least** optimised configuration, because it has to hold by construction rather than because the inliner removed it: storage is four plain Jass arrays, no hashtable native is reached for, `slotFor` takes nothing beyond the receiver and the key, and both requirements are direct calls rather than `ExecuteFunc` or a dispatch wrapper. With optimisation on, `hash(key)` becomes `key` and `equals(a, key)` becomes `a != key`.

Note the `dispatch_` functions in the output are the nullpointer check every class method gets, not type class dispatch, so the test looks at the real function underneath rather than asserting they are absent.

Full test suite green.


Frotty:
Fixed — a real defect, and mine: `slotFor` reports a full table by returning a slot below the instance's section, and I guarded `get`, `has` and `remove` for it but not `put`. A ninth distinct key wrote through the sentinel: an entry nothing can reach, a count of nine, and a negative index on targets less forgiving than Lua.

`puttingIntoAFullMapChangesNothing` fills the map exactly — keys 1 to 8 hash to the eight slots — then puts a ninth and checks the size and the existing entries are untouched.

One thing the test deliberately does not assert: reading the absent key back. `get` on a missing key of a type parameter returns a stand-in the interpreter compares as unequal to everything, which is a separate bug fixed later in this stack (#1234). Asserting it here would tie this test to a downstream change.

Also merged the updated #1231 into this branch so the diff reads against its current base.

@codex review

Frotty:
Updated from master (#1230 and #1231 are in).

On the P1 — I checked before writing anything, and the mandated regression already exists. `LuaTranslationTests` carries it in full:

- `genericOverrideChainBindsRootSlotToMostSpecificImplInLua` extracts the dispatched slot from the `FSM_currentState:*_update(` call site, asserts each of six sibling states binds that slot to its own implementation, and asserts none binds it to `NoOpState_NoOpState_update`.
- `genericOverrideChainBindsGlobalStateSlotToMostSpecificImplInLua` does the same for the global-state slot.
- `luaOutputIsDeterministicForGenericOverrideSlots` compiles the same input twice and asserts the output is byte-identical.

The repro units have `State<T:>`, `NoOpState<T:>`, `FSM<T:>`, seven `extends NoOpState<Owner>` siblings, and the states instantiated as `public constant` — every element of the AGENTS.md list, in more than the required quantity.

Those tests exercise this change directly: `sharesSemanticName` feeds the alias sets that decide exactly these bindings. They are also not hypothetical guards — an earlier attempt of mine to change the class-qualified alias was reverted precisely because `overloadedMethodsDoNotAliasInLuaDispatchTables` and `moduleProvidedOverloadedOverrideDoesNotCollapseLuaSlots` failed on it.

So the coverage you are asking for is present and green; adding a second copy alongside `FastHashMapTests` would duplicate it. The FastHashMap assertion is deliberately a different check — that a slot named after a method binds that method's implementation — which is what the specialised-name bug in this PR actually produced.

@codex review

PR #1233, opened 2026-08-15
Take a closure's instantiation from its construction on Lua

Frotty:
Stacked on #1232.

Lua specialises what it can reach from a concrete type, and it finds that type on calls carrying type arguments. A closure has no such call: it is reached through the interface it implements, which is not generic, so nothing at the call site says what the body dispatches on. Only the construction knows.

The backlog said the specialised class was built but nothing called it. That was wrong — no specialised class was being built at all, which is why this took a while to see.

## What changed

Three pieces were missing, each of which already exists for the Jass path:

1. the instantiation is collected from `ImAlloc`,
2. the member access is collected, so the write that captures the closure's environment lands on the specialised field rather than being dropped as a dead store,
3. the specialised methods are bound to the roots their originals were submethods of, and the original implementation is registered as specialised so the dispatch left in it settles instead of reaching a backend that cannot express it.

## The gate is deliberate

All three are restricted to closure-generated classes, which is the case where no call names the instantiation. Ungated, the object comes from the specialised class while its methods were bound to the erased one, and **every FastHashMap Lua test fails** — the disagreement AGENTS.md §9 describes. Widening this is a decision about Lua's erasure model, not a tweak, so I stopped at the claim I can defend.

`dispatchInsideClosureIsRejectedForLua` pinned the old diagnostic and becomes `dispatchInsideClosureLua`. `WURST_LANGUAGE.md` and `CHANGELOG.md` said "Jass only for now" and no longer do.

Full test suite green.


Frotty:
Fixed, and it was reachable exactly as described. Repro:

```wurst
function helper<Q: ToIndex>(Q x) returns int
    return Q.toIndex(x)

function foo<Q: ToIndex>(Q x) returns int
    Producer p = () -> helper(x)
    return p.produce()
```

fails on the previous commit with `Type class dispatch of toIndex could not be resolved for the Lua target`. Kept as `dispatchInsideClosureThroughHelperLua`.

The gate now uses the call-aware check that already exists for functions. I left `classNeedsSpecialization` itself alone rather than widening it, since it also gates the `ImAlloc` branch of `functionNeedsSpecialization` — widening there would change what gets specialised on paths that have nothing to do with closures, and this PR's whole safety argument rests on that gate staying narrow.

Specialising the closure class then surfaced a second problem your suggestion would have hit too: the member access collector runs again for the copy, and a receiver still typed by the generic class cannot adapt to the specialised one, which is not a superclass of it — `Could not adapt receiver Producer_test<integer{toIndex}> to superclass Producer_test_specialized_integer`. A class with no type variables left has nothing to select, so it is skipped now.

TypeClassTests, FastHashMapTests, GenericsWithTypeclassesTests, ClosureTests, LuaTranslationTests, FieldIterationTests and GenericsTests all pass.

@codex review

Frotty:
Updated from master, both addressed.

**Nested closures (P1) — real, reproduced, fixed.** Your repro gives `Rewrite generics with wrong sizes / generics: <> / typevars: [Q_captured837]`. Worth being precise about severity though: I checked the same program on master, and it fails there too — with `could not be resolved for the Lua target`. So this never compiled on Lua and is not a regression in what works; what regressed is the message, from a clean diagnostic to an internal error, which is still worth fixing.

Nested closures are now kept out of the gate, which restores the previous wording, and `nestedClosuresInsideBoundedGenericAreRejectedForLua` pins it. Making them work is a further step: the inner closure reaches its captured environment through a receiver belonging to the outer one, which has already been specialised, so specialising the owner again with what is left over is the wrong shape. That is the same reason the gate is narrow everywhere else in this PR.

**Emitted shape (P1) — taken.** `dispatchInsideClosureLua` now extracts the dispatched produce slot from the call site and asserts the allocation names the specialised class, that the specialised class binds that slot to its own implementation, and that it does not bind it to the generic original. You are right that running alone did not distinguish those, which is exactly what this change alters.

TypeClassTests, ClosureTests, GenericsWithTypeclassesTests, FastHashMapTests and LuaTranslationTests all pass.

@codex review

PR #1234, opened 2026-08-15
Resolve a type parameter's default where the binding is known

Frotty:
Stacked on #1233.

Reading a slot of a `T array` that was never written gave a stand-in rather than the default of what `T` stands for, and the stand-in compares equal only to another stand-in. So `Box<int>.first() == 0` was quietly **false** on the interpreter, while both backends had it right. Wrong answer, no error.

    class Box<T:>
        private static T array none
        static function first() returns T
            return none[0]
    init
        if Box<int>.first() == 0
            testSuccess()

The stand-in is produced by a static attribute, which cannot see the stack frames that know the type argument. `ProgramState` can, so the value is resolved where it is produced — at the array read and the member read — rather than at the comparison, which is only where the symptom shows.

Comparing an unresolved one is now an error naming the type parameter, instead of answering "not equal". The whole suite passes with it throwing, which says nothing reachable produces one any more; if something starts to, it says so rather than returning a wrong answer.

Also carries a note recording why the junk dispatch slot per specialised class stays for now: both available name sources are wrong in opposite directions, and the slot is unused, so the cost of leaving it is one table key per specialised class while the cost of getting the alias wrong is a real mis-binding.

Found while building the FastHashMap tombstone fixture, which needed a "no value" for `V`.

Full test suite green.


Frotty:
Fixed. You were right that it was half a check: `EQ` and `NOTEQ` ask the left operand, so `unresolved == 0` complained while `0 == unresolved` went quietly false — which is exactly the failure mode the refusal was added to remove.

Moved it into `WurstOperator`, where both operands are in hand, and `ILconstUnsafeDefault.isEqualTo` goes back to answering plainly. Two stand-ins still compare equal to each other; only a comparison against a real value is refused. `unwrittenArrayOfATypeParameterReadsAsItsDefaultReversed` covers the reversed order.

Also merged the updated #1233 through, keeping both tests that landed at the same spot in `TypeClassTests`.

TypeClassTests, InterpreterTests and CompiletimeTests pass; the full suite is running and I will report here if it disagrees, since this sits on the path every interpreter comparison takes.

@codex review

Frotty:
Updated from master (#1230 through #1233 are in). No new findings since the last round, and the operand-order fix is unchanged by the merge: the refusal lives in `WurstOperator` where both operands are in hand, and `ILconstUnsafeDefault.isEqualTo` answers plainly.

Merge resolution, since several files needed a side picked: took master's `EliminateGenerics`, `LuaRunnerTests` and `WurstScriptTest` (they carry the merged #1231 and #1233 work), kept ours for `BACKLOG.md`, and rebuilt `TypeClassTests` as master's version plus this PR's two `unwrittenArrayOfATypeParameter` tests so nothing from either side was dropped.

TypeClassTests, InterpreterTests and CompiletimeTests pass.

@codex review

PR #1235, opened 2026-08-15
Make generated Jass comparable across runs, and cut the suite from 13 to 7 minutes

Frotty:
Stacked on #1234.

## Determinism

The counters naming generated temporaries were per thread and never reset, so a name depended on how much had been compiled before it in the process: the same source emitted `temp0` on its own and `temp70` after other tests had run — measured, not assumed. That is what made generated Jass impossible to compare across runs, which is the cheapest way there is to check that a change did not alter the output.

They restart at the top of each compilation rather than at each flattening, because flattening happens again after every optimisation and restarting there would give one function two locals of the same name.

## Suite time

The suite ran one test at a time on an eight core machine: 786s of test time, 13m11s of wall clock. Measured from the result XML, the top ten classes are 61% of it.

Two things were needed for forking to actually help:

- **Forks rather than threads**, because the harness keeps state in statics (current test environment, global caches, extracted lua binaries) and a fork gets its own copy. Gradle hands out whole classes, so `DeterministicChecks`, which writes fixed file names, stays inside one fork.
- **A processor count per fork.** Each one otherwise sizes its GC and compiler threads for the whole machine. Eight forks doing that made every test ~3× slower and gave back only a third of the parallelism (8m34s). Capping them takes it to 7m03s. Four forks was measured too and is worse (8m39s) — less contention, but not enough parallelism to pay for it.

`ExportToWurstTest` was the slowest class at 108s, and none of it was the object data work: 30 of its 32 tests take no measurable time. Two tests looped over five cases and compiled each separately against the whole standard library. The cases declare different objects and only the compile needs the library, so they are compiled together: 108s → 67s. That class is worth the attention because wall clock cannot fall below the slowest single class.

## Also included

Two pinned diagnostics with their diagnoses, no behaviour change: the constructor case for Lua, and where subclassing a bounded generic stops. Both are recorded in `BACKLOG.md` with what a fix would have to do — including one approach I tried and reverted, because the callee has no type variables of its own so annotating the call does nothing; the call must be redirected to the specialised copy.

Full test suite green.


Frotty:
Fixed. The concern is real and the CI path is exactly as you describe — `.github/workflows/build.yml` runs plain `./gradlew test`, and `gradle.properties` gives the daemon `-Xmx3g` on top of the workers.

The count is now the smaller of half the cores and what memory allows, reserving the daemon's heap plus room for the lua and pjass processes the tests spawn. `-PtestForks=N` overrides it, and a machine that will not report physical memory falls back to cores capped at four rather than guessing high.

Kept the cores-based side as the upper bound rather than lowering it, because the measurements went the other way on this eight-core machine: serial 13m11s, four forks 8m39s, eight forks 7m03s. More forks kept winning even though each test runs about 2.6x slower under the contention — so the fix is to stop memory-poor machines over-committing, not to reduce parallelism everywhere.

Verified both paths: default resolves and runs, and `-PtestForks=2` takes effect.

Also merged #1234 through, keeping both tests that landed at the same place in `TypeClassTests`.

@codex review

Frotty:
Updated from master and the P1 is fixed — a real hazard, and one this branch introduced by forking at all.

`StdLib` guards the fetch with a `synchronized` method and an `isInitialized` flag, both of which stop guarding anything once the workers are separate JVMs pointed at the same `./temp/WurstStdlib2`. It did not bite here only because that directory already existed.

The build now fetches it once from its own JVM, before the workers start, via a task the `test` task depends on. The repository URL and pinned commit stay defined in `StdLib` rather than being repeated in the build file.

One detail worth flagging, because I got it wrong first: the task deliberately declares **no** output. Declaring the checkout as one lets Gradle skip the fetch whenever the directory merely exists — which is precisely when it may be at the wrong commit or half cloned. My first attempt did that and left the checkout sitting at the wrong revision. The fetch verifies and repairs the pinned commit itself, and costs nothing when there is nothing to do.

Verified from no checkout at all: clones, reports `Wrong version ... executing git pull`, and lands on `16729fa`.

Also fixed the logging in that path, which printed `e.getStackTrace().toString()` — the identity of an array — and told me nothing while I was diagnosing this.

DeterministicChecks, TypeClassTests and ExportToWurstTest pass.

@codex review

Frotty:
Fixed, and reproduced exactly as you described.

The repair path went by way of `master`, and the checkout this leaves behind is detached with no local `master` — so on any machine that already had a checkout, it asked for a ref that was not there. `Nothing to fetch` and `Ref master cannot be resolved` were both on that path. It now fetches from the remote with an explicit refspec and checks out the pinned commit directly, with no branch in between.

Your framing of the severity is the part I had missed: this was survivable while only the tests using the library were affected, and it is not now that every test waits on `ensureStdLib`. That amplification is this branch's doing, so a directory that cannot be opened as a repository, or has no HEAD — what an interrupted clone leaves — is discarded and cloned again rather than trusted as a cache.

Failures also print to the console now. The build gates on this, so the reason belongs in the run output; diagnosing this started from an exception that said only that something went wrong.

Checked against all four states: no checkout at all, a good one, one detached with no local `master`, and one with its `.git` removed. All four end at the pinned commit.

@codex review

Frotty:
Valid, with a caveat on how often it bites.

The structure is exactly as you describe: the `clean` and `checkout` sat outside the version check, so they ran on the first call in every process — and every worker is a process. Eight of them rewriting one directory, each taking JGit's index lock, while others read the same files.

Where I would temper it: I have not seen it fail. Three full runs at eight forks passed, because when the checkout is already correct `clean` finds nothing and `checkout` is a no-op — they still take the lock, so the window is real but narrow. It is the shape that shows up as occasional CI flakiness rather than a reproducible failure, which is reason enough given the fix is small.

Workers now read the status — which writes nothing — and only touch the directory if something is actually wrong.

Asking first turned up something else: the reset did not do what its comment claimed. Checking out the commit HEAD already points at is a no-op, so `// reset all possible changes` removed untracked files but left a modified tracked file modified. I found that by dirtying the checkout deliberately to test the repair path, and watching it not repair. It resets hard to the pinned commit now. Verified: modified file restored, untracked file removed, HEAD unchanged.

LuaTranslationTests, InterpreterTests, ExportToWurstTest and DeterministicChecks pass together, which is the parallel path this affects.

@codex review

Frotty:
Both valid, both fixed.

**Ignored artifacts.** Correct, and I confirmed the consequence rather than assuming it: `addLibDir` recurses into every subdirectory and does `getLibs().put(libName, f)` for each `.wurst` it finds, so a stale `_build` really can take a pinned package's place — quietly. And WurstStdlib2 does ignore `/_build`, `objectEditingOutput/` and `wurst.dependencies`.

Checking `getIgnoredNotInIndex()` in the worker would have reintroduced exactly what the guard was avoiding, since those artefacts appear during a run and removing them mid-run is the contention in question. So the deep clean moved to the prefetch, which runs alone before the workers — the one moment rewriting that directory disturbs nobody. Workers still repair a wrong commit or modified tracked files, which is what a run outside the build sees. Verified: a planted `_build/Stale.wurst` is removed by the prefetch and left alone by workers.

**Focused runs offline.** Also correct, and it is a documented workflow, so: `-PskipStdLibFetch` skips the task. I had to read the property at configuration time rather than in `onlyIf` — referencing `project` there fails under the configuration cache, which this build enables.

LuaTranslationTests, InterpreterTests, ExportToWurstTest and DeterministicChecks pass together, and the checkout is clean afterwards.

@codex review

PR #1237, opened 2026-08-16
Give a super call the type arguments of the class it reaches

Frotty:
A bounded generic class could not be subclassed. `class SubBox extends Box<int>` with an override calling `super.size(extra)` failed on Jass with `Typevar dispatch not eliminated`, even though the specialised copy of the superclass was built correctly next to it.

## Cause

`moveFunctionsOutOfClass` lifts a class's type variables onto each function it moves out, so `Box.size` becomes a function with its own type variable `K`. `addMemberTypeArguments` gives those type arguments back to calls that go through a receiver — `ImMethodCall` and `ImMemberAccess` — by adapting the receiver's type to the owning class.

A call that names its target outright has no receiver node to read. `super.size(extra)` and `super(k)` are both of that kind, so they were left asking for a function with type variables while supplying none. Nothing specialised them, and the super call kept pointing at the original while its dispatch went unresolved to the backend.

The error named what was inside the function rather than the reference that kept it alive, because Jass emits whatever is called rather than only what the program still holds — which is why this read as a missed dispatch rather than a missed call.

## Fix

`addReceiverTypeArguments` handles the remaining case: the receiver is still there as the first argument, so the class it is used as gives the same type arguments the receiver would have. It only touches calls that supply no type arguments to a callee that has type variables — a state that is broken however it arose. `functionOwners` records which class each function was moved out of, which is what the first attempt at this was missing.

Both super calls are covered, so the super constructor call reaches the specialised constructor too.

## Note on the previous analysis

`BACKLOG.md` claimed annotating the call could not work, on the grounds that the callee had no type variables of its own and that the call had to be redirected to the specialised copy instead. That was wrong — the lift gives it type variables — and the entry is corrected here.

## Tests

`TypeClassTests.subclassOfBoundedGenericIsRejected` becomes `subclassOfBoundedGeneric` and now executes the program. Full suite green.

The Lua half is still open and stays in the backlog: `transformGenericNewOnly` runs neither `simplifyClasses` nor `addMemberTypeArguments`, so the type variables are never lifted there, and the remaining gap is the erasure-model question tracked as item 5.

### DIFF
--- a/BACKLOG.md
+++ b/BACKLOG.md
@@ -124,6 +124,9 @@ because `LOOP.md` refers to items by number.
     never gets that table. That half is the erasure question again, as in item 5, and the Jass fix
     does not reach it: `transformGenericNewOnly` runs neither `simplifyClasses` nor
     `addMemberTypeArguments`, so on Lua the type variables are never lifted in the first place.
+    Pinned by `TypeClassTests.subclassOfBoundedGenericIsStillBrokenOnLua`, which asserts the failure
+    rather than leaving the difference between the targets to be discovered. Fixing this half makes
+    that test fail, which is the point: it then becomes a second success case.
 
 12. **Standing item, never finished.** When nothing above is left, find the next thing worth
     doing and add it here rather than stopping. Good sources, in order: a test that would have
--- a/BACKLOG.md
+++ b/BACKLOG.md
@@ -72,8 +72,8 @@ because `LOOP.md` refers to items by number.
    source parameter. Making the node canonical lets all three compare by identity and removes
    a class of silent wrong dispatch. Mechanical, well covered by the suite.
 
-13. **A bounded generic class cannot be subclassed.** Diagnosed on the Jass side and pinned by
-    `TypeClassTests.subclassOfBoundedGenericIsRejected`; the Lua half is still open. The program:
+13. **A bounded generic class cannot be subclassed on Lua.** Fixed on Jass and pinned by
+    `TypeClassTests.subclassOfBoundedGeneric`; the Lua half is still open. The program:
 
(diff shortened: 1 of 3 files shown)


Frotty:
Both addressed.

**Preserve existing method type arguments on super calls** — correct, and it was a crash rather than a missed case. The lift prepends the class's type variables to the function's own, so a call is short of that *prefix*, not of everything; my guard treated any non-empty argument list as already handled. It now compares counts and prepends only when what is missing is exactly the owning class's arity, with the adapted class type's arity checked to match before anything is added.

`superCallToAGenericMethodOfABoundedGenericClass` covers it: `Box<K: Show>` with `choose<Q>`, called as `super.choose<Marker>(new Marker(), extra)`. `K` is `int` and `Q` is a class, so the test also fails if the prefix ever goes on in the wrong order rather than only if it is missing.

**Keep bounded-subclass behavior aligned on Lua** — fair, the backlog entry alone was not enough. `subclassOfBoundedGenericIsStillBrokenOnLua` now runs the same program on Lua and pins that it compiles, runs, and never reaches `testSuccess`. It cannot use `expectError` the way `dispatchInsideConstructorIsRejectedForLua` does, because nothing errors — the object is allocated from the erased `Box` table while the specialised one holds the method — so it asserts the failure explicitly and will break as soon as that half starts working.

The Lua half stays open for the reason in item 13: `transformGenericNewOnly` runs neither `simplifyClasses` nor `addMemberTypeArguments`, so the type variables are never lifted there and a super call has nothing to carry. Closing it is the erasure-model question tracked as item 5, not something to fold into this change.

Full suite green.

@codex review

PR #1238, opened 2026-08-16
Hold an interpreted string as bytes, the way the game does

Frotty:
Warcraft III counts and indexes a string in **bytes**: `StringLength` returns a byte count and `SubString` takes byte offsets, so a slice may stop between the bytes of one character. Lua agrees, its strings being byte arrays. The interpreter held a Java string and counted UTF-16 code units, which gives the same answer only for ascii — `StringProvider` was `string.getVal().length()` and `str.substring(s, e)`.

## Why it matters now

The standard library depends on the difference rather than avoiding it. `String.wurst` sets `ENABLE_MULTIBYTE_SUPPORT = true` and:

- takes `"ä".substring(0, 1)` to find out how the engine represents half a character,
- slices a 64 character Cyrillic literal byte by byte to enumerate every continuation byte `0x80`–`0xBF`, keeping them apart by hash,
- slices an emoji for the 4-byte lead case.

Under UTF-16 the first of those returns the whole character, so `PARTIAL_CHAR_DETECTABLE` goes false and the detection concludes the engine has no multibyte characters. That is the graceful degradation its author intended for a future engine change, not a compiletime semantics we should ship: `ChunkedString` and object editor text are built from lengths the game will not agree with, and nothing reports it.

## The representation

The value is held one char per byte, so every char is below 256 and Java's own `length` and `substring` already give the game's answers. Text is encoded coming in (`fromText`) and decoded going back out to a file, a program literal, or a screen (`text()`). A half character has no text to decode to, which is the point — it keeps its byte until the other half is added back.

The constructor is private, so every one of the 21 construction sites had to be classified as text or as bytes rather than left to inspection.

Two consequences of the representation:

- **`StringCase` folds only ascii letters.** The bytes of a multibyte character are not letters; folding one the way a latin-1 char folds rewrites the character into a different one.
- **`StringHash` is computed over the bytes.** The library's `StringHash.hash` takes text and does `getBytes("UTF-8")` itself, so it cannot hash half a character, and its `byte[]` overload is private. Decoding first is not a way out either: every partial slice would decode to the same replacement character and collapse onto one hash, taking the 64 continuation bytes with it — exactly the thing the standard library tells apart. `Wc3StringHash` implements the same function (Bob Jenkins' lookup2) over bytes.

## Test runtime

The Lua runtime never implemented `StringLength` or `SubString`. The generated fallback for an undefined native raises an error, so the Lua half of any test using them could not have been passing for the reason it appeared to. Both are in `wc3shim.lua` now, as the plain Lua byte operations the game performs.

## Tests

- `StringByteSemanticsTests` — length of a two byte character, a slice cutting a character in half, and the halves rejoining into the original. Each runs on the interpreter and on Lua, so the two are pinned against each other rather than against an assumption.
- `Wc3StringHashTest` — the byte hash against the library's across ascii of every length up to 40, strings needing case and slash normalisation, and whole multibyte text, where both are defined. Then the two properties the library cannot express: the halves of a character hash apart, and all 64 continuation bytes stay distinct.

Full suite green.

## Not covered

The engine collapses every half-character slice to one marker string with a constant hash. This does not emulate that quirk: byte-accurate, a `0xD0` lead byte hashes as itself rather than matching `PARTIAL_CHAR_HASH`. The standard library still gets right answers, by a different route — `isCharBoundary` falls through to the continuation byte table, does not find a lead byte there, and reports a boundary — but `PARTIAL_CHAR_DETECTABLE` is true at compiletime while matching a narrower set of slices than in game. Reproducing the collapse is a deliberate further step and wants its own test.

### DIFF
--- a/de.peeeq.wurstscript/src/main/java/de/peeeq/wurstio/CompiletimeFunctionRunner.java
+++ b/de.peeeq.wurstscript/src/main/java/de/peeeq/wurstio/CompiletimeFunctionRunner.java
@@ -421,6 +421,25 @@ public class CompiletimeFunctionRunner implements AutoCloseable {
         return constantToExpr(trace, value, null);
     }
 
+    /**
+     * The text for a string a compiletime expression produced, which becomes a literal in the
+     * generated script.
+     * <p>
+     * A string held by the interpreter is a sequence of bytes and may hold half of a character -
+     * slicing one in half is a thing the standard library does deliberately. A literal cannot: the
+     * script is written as UTF-8 and neither Jass nor the escaping here can write a byte down
+     * numerically, so half a character would go in as the replacement character and come back out
+     * three bytes long. Refused rather than carried across at a different length.
+     */
+    private String literalText(ILconstString value, Element trace) {
+        if (!value.isText()) {
+            throw new CompileError(trace, "A compiletime expression returned a string holding part of a"
+                + " multibyte character, which cannot be written into the generated script. Slice it"
+                + " where the program runs rather than at compiletime, or keep whole characters.");
+        }
+        return value.text();
+    }
+
     private ImExpr constantToExpr(Element trace, ILconst value, @Nullable ImType expectedType) {
         if (value instanceof ILconstBool) {
             return JassIm.ImBoolVal(((ILconstBool) value).getVal());
@@ -429,7 +448,7 @@ public class CompiletimeFunctionRunner implements AutoCloseable {
         } else if (value instanceof ILconstReal) {
             return JassIm.ImRealVal("" + ((ILconstReal) value).getVal());
         } else if (value instanceof ILconstString) {
-            return JassIm.ImStringVal(((ILconstString) value).text());
+            return JassIm.ImStringVal(literalText((ILconstString) value, trace));
         } else if (value instanceof ILconstNull) {
             return expectedType == null ? ImHelper.nullExpr() : JassIm.ImNull(expectedType.copy());
         } else if (value instanceof ILconstTuple) {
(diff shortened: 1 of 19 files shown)


Frotty:
Both addressed.

**Decode interpreter strings before passing them to SQLite** — my mistake, and exactly as described: I converted the inbound read and left the outbound side on `getVal()`. `sqlite_open`, `sqlite_prepare`, `sqlite_bind_string` and `sqlite_exec` all decode now.

`CompiletimeNativesTest.sqliteRoundTripsNonAsciiText` binds `"Grüße 日本"`, reads it back, and checks both the text and the bytes match, so a length taken either side of the round trip agrees. It also runs a non-ASCII literal in the SQL itself, which takes the `sqlite_exec`/`sqlite_prepare` path. It fails without the fix.

**Preserve partial bytes when materializing compiletime strings** — the diagnosis is right and the test gap was real, but lossless materialisation is not reachable from here. The value has to become an `ImStringVal` that both backends print through `Utils.escapeString`, and the script is written as UTF-8 (`WurstCompilerJassImpl:658`). That escaper emits only `\n \r \t \" \`, and Jass has no numeric escape either — so a lone `0xC3` has no representation in a generated literal. Writing it as a char yields `0xC3 0x83`, which is the same corruption wearing different clothes. Making it representable means byte strings through the IM and both writers, which is a much larger change than this one.

So it refuses rather than mangles: `literalText` checks the bytes round-trip through text and raises a `CompileError` naming the situation when they do not.

Two tests at that boundary, since it was untested either way:

- `aCompiletimeStringKeepsItsLengthAfterTransforms` — `compiletime("hällo")` is still six bytes after transforms and still equals the literal, so whole characters cross unchanged. This is the case the standard library actually relies on.
- `aCompiletimeStringHoldingHalfACharacterIsRefused` — the reviewed example, pinned as a refusal rather than as a length of 3.

Flagging the judgement call rather than burying it: refusing is not what was asked for, and if the larger byte-through-the-backend change is wanted, this is one line to swap out and the test documents precisely what would change.

Targeted tests green; full suite still running and I will report if it turns.

@codex review

PR #1239, opened 2026-08-16
Keep the fields a specialised class allocates, and compile FastHashMap against the standard library

Frotty:
Every existing `FastHashMapTests` case compiles a bare `package test` with no imports. The container is meant to live in the standard library, so this compiles it with the library present — and that immediately found a real bug.

## The bug

A specialised class allocated **an empty table**:

    function FastHashMap:create2()
        local new_inst = ({FastHashMap_base=0, FastHashMap_count=0, })   -- erased

    function FastHashMap_specialized_integer__integer:create3()
        local new_inst = ({})                                            -- specialised

A specialised class holds *copies* of the original's fields, and nothing refers to the copies: an access made before specialisation still names the original's variable. Both passes that drop unread fields — `RemoveGarbage` for Lua and `ImOptimizer` — therefore dropped every one of them. An instance allocated from the specialised class then came out with no fields while the emitted code went on reading them by name, so `size()` returned `nil` and `put` incremented `nil`.

Lua resolves a field by name and both classes carry the same one, so a name read anywhere keeps the field wherever it is allocated. The cost is an unused table key in the worst case; the alternative is an allocation that silently omits everything.

It was reachable only where an allocation goes through the specialised class rather than the erased one, which is why nothing caught it — the existing tests all allocate through the erased class. Putting the library in scope is what moved an allocation onto the other path.

## Tests

- `fastHashMapAgainstTheStandardLibrary` — the int-key map, compiled and run with the library present.
- `classKeyAgainstTheStandardLibrary` — a class key with a user written `Hashable` instance, which is the shape a real map is used in.
- `fastHashMapAgainstTheStandardLibraryLua` — compiled, and asserts a specialised class allocates the same fields as the class it was specialised from. That assertion is what fails without the fix.

Full suite green.

## Note on the Lua test being compile-only

It is compiled rather than run, as every other test putting the standard library on Lua is. No standard library program has ever executed on that target in this repository: the runtime shim cannot initialise the library's own packages, and `GameTimer` fails first. That is a harness gap unrelated to generics, worth its own issue rather than being folded in here.

### DIFF
--- a/de.peeeq.wurstscript/src/main/java/de/peeeq/wurstio/WurstCompilerJassImpl.java
+++ b/de.peeeq.wurstscript/src/main/java/de/peeeq/wurstio/WurstCompilerJassImpl.java
@@ -956,7 +956,7 @@ public class WurstCompilerJassImpl implements WurstCompiler {
             timeTaker.endPhase();
         }
         beginPhase(13, "lua remove garbage");
-        RemoveGarbage.removeGarbage(imProg);
+        RemoveGarbage.removeGarbage(imProg, imTranslator);
         imProg.flatten(imTranslator);
         timeTaker.endPhase();
 
--- a/de.peeeq.wurstscript/src/main/java/de/peeeq/wurstscript/translation/imoptimizer/ImOptimizer.java
+++ b/de.peeeq.wurstscript/src/main/java/de/peeeq/wurstscript/translation/imoptimizer/ImOptimizer.java
@@ -19,6 +19,8 @@ import de.peeeq.wurstscript.types.TypesHelper;
 import de.peeeq.wurstscript.utils.Pair;
 import de.peeeq.wurstscript.validation.TRVEHelper;
 
+import java.util.stream.Collectors;
+
 import java.util.*;
 
 public class ImOptimizer {
@@ -124,6 +126,7 @@ public class ImOptimizer {
             ImProg prog = trans.imProg();
             trans.calculateCallRelationsAndReadVariables();
             final Set<ImVar> readVars = trans.getReadVariables();
+            final Set<String> readFieldNames = readVars.stream().map(ImVar::getName).collect(Collectors.toSet());
             final Set<ImFunction> usedFuncs = trans.getUsedFunctions();
             SideEffectAnalyzer sideEffectAnalyzer = new SideEffectAnalyzer(prog);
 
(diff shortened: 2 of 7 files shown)


Frotty:
Fixed properly rather than narrowed — you are right that matching names program-wide is too blunt, and `size` is exactly the name that would bite on a standard library build.

The pairing is now recorded where it is known. `specializeClass` copies a class structurally, so field *i* of the copy is field *i* of the original; it records each pair on the `ImTranslator` it already holds. Both passes then keep a field when it is read **or** when the field it was copied from is read:

    field -> readVars.contains(field) || readVars.contains(trans.originalOfSpecializedField(field))

`originalOfSpecializedField` returns the argument unchanged for anything that is not a copy, so an ordinary dead field is dropped exactly as before — including a dead `size` in a class unrelated to any specialisation. The name check is gone from both `ImOptimizer` and `RemoveGarbage`.

`RemoveGarbage.removeGarbage` takes the translator now; its only call site (`WurstCompilerJassImpl:962`) already had one to hand.

The assertion that a specialised class allocates the same fields as the class it was specialised from still passes, so the original bug stays fixed while the retention is confined to actual copies. Full suite green.

@codex review

Frotty:
Good catch, and it was a hole in the previous fix rather than a separate issue: keeping the copy alive is worth nothing if it is alive under a different key.

Reproduced before changing anything, by renaming `FastHashMap`'s `count` field to `size` so it collides with `size()`:

    erased:      FastHashMap_base=0, FastHashMap_size_field=0,
    specialised: FastHashMap_base=0, FastHashMap_size=0,

The allocation writes one key and every access reads the other, exactly as described.

Fixed by carrying the origin relationship into normalisation. `alignSpecializedFieldNames` runs after `normalizeFieldNames` and gives every copy the name of the field it was copied from. The accesses name the original's field, so the original's name is the only key either can use — and this holds whether or not the origin's own class survived pruning, since the name is taken from the variable rather than from its class.

`aFieldNamedLikeAMethodKeepsOneKeyAcrossSpecialisation` is the reproduction above, kept as a test: it runs the map on Lua with the colliding name and asserts the specialised class allocates the same fields as the erased one. It fails without the alignment.

Full suite green.

@codex review

Frotty:
Fixed, with one correction to how I am reporting it.

The objection is right: restoring the origin's name unconditionally undoes an escape the specialisation had made for itself. The name is now settled once per field, against the method names of the origin **and** of every specialisation holding a copy, and every copy takes that name. `collectNamesEachFieldMustAvoid` gathers those names per original field before any renaming; `normalizeFieldNames` then decides a name the first time any member of the group is met and reuses it for the rest. Neither side can be renamed into the other's slot, and the two still end up as one key.

**I could not reproduce the specific failure.** I wrote a test for it first — a `FastHashMap` field named `get_specialized_integer__integer`, as described — and it passes against the previous commit as well, so it never demonstrated the bug. The reason is in the emitted Lua: nothing in `FastHashMap` overrides anything, so its specialised methods come out as direct calls (`FastHashMap_get_specialized(m, 1)`) rather than dispatch slots, and there is no method key for a field to shadow. A faithful reproduction needs a generic class with a virtual method, which is a different shape than this PR builds.

Rather than keep a test whose name claims more than it checks, I removed it. The fix rests on the reasoning above instead: it is safe by construction against both sets of method names, which is what the comment asked for.

The reproduction from the previous round is still there and still meaningful — `aFieldNamedLikeAMethodKeepsOneKeyAcrossSpecialisation` covers a field colliding with a method of its own class, which does occur here, and it fails without the fix.

Full suite green.

@codex review

PR #1241, opened 2026-08-16
Rewrite the backlog around what is left after the container works

Frotty:
Documentation only. The goal these items were organised around is met, so the Todo said things that are no longer true — item 13 read as open when its Jass half landed in #1237, and the ordering still pointed at the feature rather than at what surrounds it.

What changed:

- **Done** records what actually landed: the container works on both targets and against the standard library on Jass, the Jass half of subclassing, and byte-oriented strings with the library bump they unblocked. Each notes what was wrong in the earlier analysis, since two items here previously asserted a cause that turned out to be false.
- **Two new items for gaps in what the suite can see.** Nothing executes a standard library program on Lua (16), and a bump of the pinned library is only checked for compiling (17). Both were discovered while shipping the container, and the first is how a specialised class allocating no fields at all survived — the paths that would have caught it are compiled and never run.
- **The erasure question is one item (18)** instead of the same paragraph repeated under items 6 and 13. Both end there, and #1239 is a reason to take it seriously: two class shapes existing at once is what produced the field bugs, and while those are fixed, the shape that produced them is not.
- **The flaky determinism test is item 20**, having blocked #1240 for a run and passed on re-run of the same commit.
- Item 9 no longer names `WURST_LANGUAGE.md`, which is not in this repository.
- Item 10 gains the note that `moveFunctionsOutOfClass` copies type variables deliberately and #1237 depends on that copy, so what is being made canonical is per scope rather than per program.

### DIFF
--- a/BACKLOG.md
+++ b/BACKLOG.md
@@ -135,9 +135,12 @@ itself, and one gap in what the suite can see.
     that the test is at fault. Calling it flaky was too quick.
 
     Diagnose it instead: capture both outputs on a failing run and diff them, and rule out harness
-    interference rather than assuming it. `WurstScriptTest` clears the global caches around every
-    test method, so the two compiles do start from the same cache state; whatever differs is
-    somewhere else.
+    interference rather than assuming it. The two compiles do start from the same cache state, but
+    that comes from two separate resets: `WurstScriptTest`'s `@BeforeMethod` clears before the
+    first, and the explicit `GlobalCaches.clearAll()` between the compilations inside the test
+    clears before the second. Both are load bearing — remove either and the comparison stops being
+    between equal starting states, which would invalidate the conclusion rather than explain the
+    failure.
 
 12. **Standing item, never finished.** When nothing above is left, find the next thing worth
     doing and add it here rather than stopping. Good sources, in order: a test that would have
--- a/BACKLOG.md
+++ b/BACKLOG.md
@@ -17,7 +17,7 @@ The container these bounds were added for now compiles and runs against the stan
 Jass, and compiles on Lua (#1239). What is left is generality around it rather than the feature
 itself, and one gap in what the suite can see.
 
-16. **Nothing executes a standard library program on Lua.** Every test which puts the library on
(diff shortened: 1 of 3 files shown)


Frotty:
@codex review

Frotty:
All three addressed, and two of them were errors of mine rather than differences of opinion.

**Keep the compile-twice determinism check** — you are right, and I was too quick to call it flaky. It compiles one repro twice and compares byte for byte, so an intermittent mismatch is evidence of intermittent nondeterminism in Lua emission, which is the thing it exists to catch; a re-run passing says the nondeterminism is intermittent, not that the test is at fault. The item now says so in those terms, says not to weaken the assertion, and says to diagnose instead — capture both outputs on a failing run and diff them, and prove harness interference rather than assume it. One thing already ruled out: `WurstScriptTest` clears the global caches around every test method, so both compiles do start from the same cache state.

**Assign unused IDs** — correct, `Done` already holds 16, 17, 18 and 19. The new items are 21, 22, 23 and 24, and the cross references in items 6, 13 and 22 follow.

**Keep the tracked language digest** — also correct, and my claim that it was external was simply wrong: I searched two levels from the repository root and it sits under `src/main/resources/agent-docs`. Item 9 names it again.

Since the standing practice is to fold documentation into whichever item changes the behaviour, and two items landed without doing so, the digest now also covers what shipped: subclassing a bounded generic class and that it does not work on Lua, and strings being byte sequences on both targets, including that a compiletime expression cannot return half a character.

Also caught by the follow-up: `TypeClassTests.subclassOfBoundedGenericIsStillBrokenOnLua` pointed at item 5, which is finished. It points at item 23 now, and that is the only backlog reference in code.

@codex review

Frotty:
Right, and the distinction matters for exactly the reason given: the conclusion depends on which reset is doing the work.

The item now names both. `@BeforeMethod` clears before the first compilation, and the explicit `GlobalCaches.clearAll()` between the two compilations inside the test clears before the second; both are load bearing, and removing either stops the comparison being between equal starting states, which would invalidate the conclusion rather than explain the failure.

@codex review

PR #1242, opened 2026-08-16
Run a standard library program on Lua

Frotty:
Nothing in this repository has ever executed a standard library program on Lua. Every test putting the library on that target compiles only — which is why `fastHashMapAgainstTheStandardLibraryLua` asserted on emitted shape rather than on a result, and it is how the empty-allocation bug in #1239 survived: the paths which would have caught it are compiled and never run.

It turned out to be four natives.

## What was missing

Three packages could not initialise, each on one undefined global:

| Native | Package |
|---|---|
| `StringHash` | Colors |
| `Location` | Vectors |
| `TimerStart` | GameTimer |

`StringCase` made a fourth once the program itself ran. The wrapped message (`Could not initialize package GameTimer.`) hides this; the underlying error is on the line above it, which is where I should have looked the first time.

## The implementations

Byte-accurate rather than approximations, since a test comparing two targets is worth nothing if they disagree by construction:

- **`StringHash`** is the same Bob Jenkins lookup2 as `Wc3StringHash`, over bytes, with the same normalisation. It is a second transcription of one algorithm, which can drift, so `agreesWithTheLuaRuntimeImplementation` pins both against a fixed corpus — ascii, a path with backslashes, a multibyte character and the empty string. A change parting the two fails there.
- **`StringCase`** folds only ascii letters, matching the interpreter: the bytes of a multibyte character are not letters, and folding one rewrites the character.
- **`Location`** is a plain pair; nothing in a test reads terrain from one.
- **`TimerStart`** keeps what it was started with and never fires. A test drives its own program rather than waiting on game time, and a package which starts a timer at init only needs the call to succeed. Making timers actually fire needs a model of game time, which is a design question rather than more stubs — worth saying plainly, because this is the one stub that is not equivalent to the real thing.

## The harness

Success is read off stdout. Without the library, a program gets the generated fallback for `testSuccess`, which prints. With the library it gets the library's own, which is empty — so a test on that target could only ever be reported as not having succeeded, whatever it actually did. The harness now installs a printing one after loading the script.

This is the part that matters beyond the natives: it was not possible for a standard library Lua test to pass.

## Result

`fastHashMapAgainstTheStandardLibraryLua` executes now instead of being compiled and assumed, and the container works there. Full suite green.

Closes the executed-on-Lua half of backlog item 16 (#1241).

### DIFF
--- a/de.peeeq.wurstscript/src/test/java/tests/wurstscript/tests/FastHashMapTests.java
+++ b/de.peeeq.wurstscript/src/test/java/tests/wurstscript/tests/FastHashMapTests.java
@@ -413,7 +413,7 @@ public class FastHashMapTests extends WurstScriptTest {
      */
     @Test
     public void fastHashMapAgainstTheStandardLibraryLua() throws IOException {
-        test().withStdLib().testLua(true)
+        test().withStdLib().testLua(true).executeProg()
             .lines(withStandardLibrary(program(fastHashMap(), INT_INSTANCE, USE_WITH_COLLISION)));
         assertSpecialisedClassesAllocateTheirFields(
             Files.toString(new File("test-output/lua/FastHashMapTests_fastHashMapAgainstTheStandardLibraryLua.lua"),
--- a/de.peeeq.wurstscript/src/test/java/tests/wurstscript/tests/Wc3StringHashTest.java
+++ b/de.peeeq.wurstscript/src/test/java/tests/wurstscript/tests/Wc3StringHashTest.java
@@ -15,7 +15,7 @@ import static org.testng.Assert.assertNotEquals;
  * are only comparable where the library is defined - a string of whole characters - so that is what
  * is compared, and it is enough: the arithmetic is the same for every input, only the bytes differ.
  */
-public class Wc3StringHashTest {
+public class Wc3StringHashTest extends WurstScriptTest {
 
     private static void agrees(String text) throws UnsupportedEncodingException {
         assertEquals(Wc3StringHash.hash(ILconstString.fromText(text).getVal()),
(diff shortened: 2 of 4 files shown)


Frotty:
Both fixed.

**Preserve termination in the injected success native** — right, and it is a real behaviour difference rather than a cosmetic one. `LuaNatives` emits `print("testSuccess") os.exit()` and the interpreter's `OutputProvider.testSuccess()` throws its sentinel, so both stop the program. My replacement printed and returned, which would newly fail or time out a test that succeeds and then runs into something that fails or loops. It calls `os.exit()` now, matching the generated native.

**Exercise the Lua hash in the parity test** — also right, and the test was misnamed for what it did: every assertion ran the Java implementation against numbers I had pasted in from a Lua run, so a change breaking only the Lua side left it green. That is the case worth catching, and it was the one case it could not catch.

It now runs each input through `wc3shim.lua` with the bundled interpreter and compares the two implementations directly. The input is escaped byte by byte into the Lua literal, so what the shim hashes is what the Java side was handed rather than whatever the command line did to it on the way through. Inputs cover ascii, both slash forms, mixed case, a two byte character, a three byte one and the empty string.

Full suite green.

@codex review

Frotty:
Fixed. `luaHashOf` catches the `IllegalStateException` from `getLuaExecutable()` and raises `SkipException` with the reason, so a host without a working interpreter skips visibly rather than failing the class — the same handling as the normal execution path, and nothing in this test is what such a host would be exercising anyway.

One thing to flag about the history rather than leave it to be noticed: commit f78ca81 carries the wrong message. Its content is this skip handling; the message belongs to a change on the backlog branch, and a `checkout` blocked by an uncommitted file put the two together. The content of every commit here is what it claims to do, but that one's description is not. It squashes away on merge, so master will carry this pull request's title instead — saying so because a reader of the branch would otherwise be misled.

Frotty:
@codex review

Frotty:
Right, and it is the same bug the main Lua harness had — which I fixed earlier in this work and then reintroduced here, in a helper small enough that it did not look like it needed the same care.

`r.lines().collect(...)` waits for the stream to close, and a hung process never closes it, so execution never reached the `waitFor`. The 30 seconds could only have elapsed after the hang had already stopped the suite, which is precisely the failure it was there to bound.

Now: drained on a thread of its own, waited for with the deadline, `destroyForcibly()` when it expires, and the failure names the input rather than leaving a bare timeout. The drain thread is joined afterwards so the output is fully written before it is read.

Parity test still passes on all eight inputs.

@codex review

PR #1244, opened 2026-08-16
Find a type variable by what it was copied from, not by its name

Frotty:
`EliminateGenerics` found a type variable's position by comparing names when identity failed. The backlog called this a name-tolerant lookup compensating for several nodes standing for one source parameter, and called removing it mechanical. Neither was quite right.

## What it actually was

Removing the name comparison broke most of the generics tests, not an edge case — so identity never held on the common path. Tracing which nodes failed to match, the target was always **detached**: a type variable belonging to nothing.

Three places copy a function's or a class's type variables and then empty the copy's list:

- `specializeFunction`
- the specialised method implementation beside it
- `specializeClass`

`copyWithRefs` copies the type variables along with the body, so the references inside the copy point at the new nodes; emptying the list then leaves those nodes with no owner. The name comparison was not compensating for the occasional duplicate — it was the only thing holding those references together.

## The change

Each copy is recorded against what it was made from, before the list is emptied, and the lookup compares those. Same shape as the field origin map in #1239, and for the same reason: the relationship is known where the copy is made and nowhere afterwards.

After pairing all three sites, no lookup falls back to a name, so the comparison is gone from `indexOfTypeVar` and from `inheritTypeClassBinding` through it. Two parameters which merely share a name can no longer be taken for one where dispatch is resolved.

## What is left

`ProgramState.getCurrentTypeArgument` still matches on names. The record lives on the `ImTranslator`, and the interpreter is handed a program rather than the translation which produced it, so reaching it means threading the translator through the interpreter — four construction sites, three of them tests, but an API change all the same. The comment there now says that, rather than leaving it looking like an oversight, and backlog item 10 carries what remains.

## On the test

`aMethodTypeParameterMayShareTheClassParameterName` covers a class and a method declaring a parameter of the same name. **It passes without this change as well as with it**, so it does not pin the fix — the lookups are not reached with both parameters in one list by that shape, and I did not find a program which reaches them that way. It is kept for the language case, and its comment says as much. The change rests on removing the name comparison and on the suite showing nothing depended on it.

Full suite green, 7m41s.

### DIFF
--- a/BACKLOG.md
+++ b/BACKLOG.md
@@ -87,15 +87,26 @@ itself, and one gap in what the suite can see.
    the repository root, which is easy to miss when looking for it. Fold an update into whichever
    item changes the behaviour rather than doing it as a separate pass.
 
-10. **One `ImTypeVar` per type parameter.** Name-tolerant lookups remain in
-   `EliminateGenerics.indexOfTypeVar`, `inheritTypeClassBinding` and
-   `ProgramState.getCurrentTypeArgument`, compensating for several nodes standing for one
-   source parameter. Making the node canonical lets all three compare by identity and removes
-   a class of silent wrong dispatch. Mechanical, well covered by the suite.
+10. **The interpreter still finds a type argument by name.**
+    `ProgramState.getCurrentTypeArgument` compares `ImTypeVar` nodes by name, so two parameters which
+    merely share one look like the same parameter.
 
-   Note before starting: `moveFunctionsOutOfClass` copies a class's type variables onto each
-   function it moves out, deliberately, and #1237 depends on that copy. Identity cannot hold across
-   that boundary, so what is being made canonical is per scope rather than per program.
+    The two lookups in `EliminateGenerics` no longer do. What they compare is what each node was
+    copied from: three places copy a function's or a class's type variables and then empty the copy's
+    list, which leaves every reference inside pointing at a node belonging to nothing — the name
+    match was not compensating for the odd duplicate, it was the only thing holding those references
+    together. Recording the copy before the list is emptied replaces it entirely, and name matching
+    is gone from `indexOfTypeVar` and from `inheritTypeClassBinding` through it.
+
+    The record lives on the `ImTranslator`. The interpreter is handed a program rather than the
+    translation which produced it, so reaching the record from `ProgramState` means threading the
+    translator through the interpreter — four construction sites, three of them tests, but an API
+    change all the same. That is what is left of this item.
+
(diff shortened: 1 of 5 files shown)


Frotty:
You were right, and the recipe was the missing piece. `aBoundedMethodParameterMayShareTheClassParameterName` now fails without the change.

What my version was missing: the method's parameter was unbounded, so nothing ever dispatched on it and `indexOfTypeVar` was never asked to tell it from the class's. Bounding it and giving it a **different** instance is what makes both parameters reach the same lookup.

    class Holder<T: Show>          // built over int, whose instance doubles
        function shown() returns int
            return T.show(held)
        function convert<T: Show>(T other) returns int   // called with string, whose instance answers 7
            return T.show(other)

With the change, `h.shown() == 42` and `h.convert<string>("x") == 7`. Without it:

    ClassCastException: ILconstString cannot be cast to ILconstNum

The string argument is dispatched through the int instance, which does `x * 2`. So this was a wrong dispatch rather than the hygiene cleanup the backlog item described — worth saying, because I had accepted that framing and it made the change look less necessary than it is.

I have also corrected the item's note, which claimed a same-named parameter test passes either way. That was true of the weak version and misleading as a general statement; it now records that the bound and the second instance are what bring the two parameters into one lookup.

Full suite green, 7m48s.

@codex review

Frotty:
Tried, and Lua coverage of this fix cannot be given. What I found instead is a gap worth recording.

**The shape does not compile on Lua.** The only program which brings two same-named parameters into one lookup is a class parameter beside a *method* parameter, and that is rejected on Lua with `Generics should match class method type variables` — on master as well as here, so it is pre-existing rather than something this change causes.

**A Lua-compatible variant covers nothing.** I wrote one with the second parameter on a free function rather than a method. It compiles and runs on Lua, and it passes *with the change and without it*: the two parameters never reach one lookup that way. I checked it against master before believing it, and deleted it rather than ship a test whose name claims more than it checks — the same mistake as the version you caught.

So the Lua behaviour is pinned as a rejection, `aBoundedMethodParameterInAGenericClassIsRejectedForLua`, with the reason in its comment. Should the gap be fixed, that test fails and gains the dispatch assertion the interpreter one already has.

The gap is **backlog item 25**, with where to start: the message comes from the arity check between a call's generics and the callee's type variables. On the Jass path a method of a generic class has the class's variables lifted onto it; `transformGenericNewOnly` does not lift them, so the method's own parameter is counted against a list which does not include the class's.

The interpreter repro from the previous round is unchanged and still fails without the fix, with the `ClassCastException` from dispatching a string through the int instance.

Full suite green, 8m.

@codex review

PR #1245, opened 2026-08-16
Make the determinism failure say what differed

Frotty:
`luaOutputIsDeterministicForGenericOverrideSlots` compiles one repro twice and compares the output byte for byte. It failed once on Windows CI, passed on a re-run of the same commit, and blocked an unrelated pull request in between — producing nothing anyone could act on, because comparing two whole scripts with `assertEquals` prints an unreadable dump and the outputs are gone by the time the failure is read.

## Not reproduced

Worth saying first, since it shapes what this change is.

250 compiles of that repro in one JVM, caches cleared between each, came out byte-identical. I also read the places in the emission path where hash-ordered iteration could reach the output, rather than guessing at them:

- `TypeId.calculate` sorts by class name and package
- `createMethods` groups through a `TreeMap`
- `assignDispatchAliases` collects into a `TreeSet` and iterates a list
- `collectSuperClasses` uses its set only to mark what it has already seen

None of those can vary between runs. So whatever differs is either rarer than one in 250, or comes from something a local run does not vary — a different core count changing the fork layout, memory pressure, or that runner's interpreter build.

## What changed

The assertion is untouched. It still compiles twice and still compares byte for byte, because an intermittent mismatch is evidence of nondeterminism in Lua emission and that is what it exists to catch.

What is new is that a failure carries evidence: both scripts are written beside the test output, and the first differing lines are named with their numbers. If the scripts differ but no line does, it says that too, since then it is line endings or trailing bytes rather than emission.

The next occurrence will say what differed instead of costing a re-run to dismiss.

Backlog item 24 records the negative result and where not to look again, so this is not re-derived from scratch next time.

Full suite green, 6m14s.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -127,6 +127,21 @@ jobs:
           check_name: Test Results (${{ matrix.os }})
           fail_on_failure: false
 
+      # luaOutputIsDeterministicForGenericOverrideSlots compiles one program twice and keeps both
+      # scripts when they differ. They are written to the runner's filesystem, so without this the
+      # evidence goes with the runner and the failure is as unactionable as it was before it kept
+      # anything. Most failures are not this test, hence if-no-files-found: ignore.
+      - name: Upload determinism scripts (on failure)
+        if: failure()
+        uses: actions/upload-artifact@v4
+        with:
+          name: determinism-scripts-${{ matrix.os }}
+          path: |
+            de.peeeq.wurstscript/test-output/determinism-first.lua
+            de.peeeq.wurstscript/test-output/determinism-second.lua
+          if-no-files-found: ignore
+          retention-days: 14
+
       - name: Upload packaged artifact (per-OS)
         uses: actions/upload-artifact@v4
         with:
--- a/BACKLOG.md
+++ b/BACKLOG.md
@@ -125,22 +125,27 @@ itself, and one gap in what the suite can see.
     this stops being dead weight, because the cost of getting it wrong is a real mis-binding while
     the cost of leaving it is one unused table key per specialised class.
 
-24. **`luaOutputIsDeterministicForGenericOverrideSlots` fails intermittently.** It failed once on
-    Windows CI and passed on a re-run of the same commit, having blocked an unrelated pull request
+24. **`luaOutputIsDeterministicForGenericOverrideSlots` failed once and has not since.** It failed
+    on Windows CI and passed on a re-run of the same commit, having blocked an unrelated pull request
     in between.
 
     Do not weaken the assertion. It compiles one repro twice and compares the output byte for byte,
-    so an intermittent mismatch is evidence of intermittent nondeterminism in Lua emission, which is
-    exactly what it exists to catch — a re-run passing says the nondeterminism is intermittent, not
-    that the test is at fault. Calling it flaky was too quick.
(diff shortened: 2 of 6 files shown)


Frotty:
Correct, and it undercut the whole point of the previous change: I said the failure keeps both scripts, which is true locally and worthless on CI. They go with the runner, and the only upload step is the release archive, which does not run on failure at all — so the evidence was still gone by the time anyone read the check.

Both halves of your suggestion, since they cover different readers:

**The scripts survive the job.** A new step uploads them when a job fails:

    - name: Upload determinism scripts (on failure)
      if: failure()
      uses: actions/upload-artifact@v4
      with:
        name: determinism-scripts-${{ matrix.os }}
        path: |
          de.peeeq.wurstscript/test-output/determinism-first.lua
          de.peeeq.wurstscript/test-output/determinism-second.lua
        if-no-files-found: ignore
        retention-days: 14

`if-no-files-found: ignore` because most failures are not this test, and the path is repository-root relative since the job's `working-directory` applies only to `run` steps.

**The report says what changed without fetching anything.** The message no longer stops at five differing lines: it states how many lines differ and lists up to forty. An artifact needs someone to go and get it; the check is what gets read first, so the message has to stand on its own.

Full suite green, 7m9s. The workflow change is indentation-checked against its sibling step, since a broken workflow would block every pull request rather than just this one.

@codex review

Frotty:
Fixed. Comparing equal indexes was not a diff, and the case it described worst — a moved or inserted block shifting everything after it — is the one this diagnostic exists to investigate.

Aligned on the longest common subsequence now. A moved block reads as one removal and one addition, the summary counts each side separately, and a pathological pair reports the two sizes rather than allocating a table for it.

`DeterminismDiffReportTest` covers the summary itself: an insertion reads as one addition rather than a cascade, a move as one removal plus one addition, an in-place replacement names both versions, and two scripts whose lines all match report the trailing-byte case. It will only ever be read during a failure nobody can reproduce, so it is worth testing rather than trusting.

One caveat on verification: I tried to run those tests against the previous implementation to confirm they fail there, made a mess of the git juggling, and got two results that were meaningless because the tree was not in the state I thought. What I can state is narrower and checkable by inspection — they assert on phrasing and counts the index version cannot produce, since its message was `"N line(s) differ"` and contains none of them. I did not get a clean execution against the old code.

Narrow tests green (`DeterminismDiffReportTest`, `LuaTranslationTests`).

@codex review

Frotty:
Two corrections since the last round, one of them mine to own.

**The line-ending case was dead code, as you said.** The caller only asks after finding the two unequal, and splitting on `"\n"` alone leaves the carriage return in the line text, so CRLF against LF differed on every line and any trailing difference read as a changed line. My test for that branch passed two identical strings — a state production never reaches — which is exactly why it looked covered.

Split on `\r?\n` now, so the two align line for line and the branch is genuinely reached. It says what it means: the difference is in terminators or trailing bytes, with both lengths. The test passes CRLF against LF, asserts the wording, and asserts that no line is reported as removed.

**I also pushed two files here that do not belong to this branch.** The dispatch slot changes for #1247 rode along on a branch switch with uncommitted changes and went out in `833023c95`, which also made this branch red — they break two override chain tests. Taken back out in `d8d3f1b5a`; this branch is the determinism diagnostic and nothing else, and the debug probe I left in the diff report test while checking its wording is gone with them.

Narrow tests green: `DeterminismDiffReportTest` and `LuaTranslationTests`.

@codex review

PR #1246, opened 2026-08-16
Pin why module bounds cannot be fixed by rewriting the receiver

Frotty:
`module M<T: Show>` is rejected today. Backlog item 7 proposed two ways to fix it — "receiver rewriting during expansion, or type parameters on `ModuleInstanciation`". I tried the first. It cannot work, and this records why so it is not rediscovered.

No behaviour change: a test and the backlog entry.

## What was tried

Expansion copies the module body into the user and replaces the module's type parameters **in type positions**. A requirement of a bound is called on the parameter itself, `T.show(x)`, and that receiver is a name resolved through `lookupType`, so the replacement never reaches it. Renaming it to the using class's parameter is what "receiver rewriting" meant.

The rename works. With it, the error moves from the blanket rejection to `Could not find variable K` at the dispatch — which is not a mistake in the rename but this, in `NameResolution.nextScope`:

    if (currentScope instanceof ModuleInstanciation) {
        return nextScope(moduleInstanciation.attrModuleOrigin());
    }

A module body resolves in the module's own scope, not in the scope of whoever uses it, so it cannot capture the user's names. That is deliberate, and it means a receiver renamed to the using class's parameter names something the scope cannot see. The approach is ruled out rather than incomplete.

Reverted, so the tree is unchanged.

## What is left

The other half of the original note: **type parameters on `ModuleInstanciation`**. The instantiation declares the parameter itself, bound to the argument, so the copied body keeps saying `T` and `T` resolves with no rename. That needs `ModuleInstanciation` to carry type parameters in `wurstscript.parseq` and everything reading that node to follow — a grammar change rather than a patch to the expander, which is why I stopped instead of half-building it.

A second case survives even then: `use Shower<int>`. A requirement is dispatched on a type parameter, so `int.show(x)` is not a dispatch at all; a concrete argument has to resolve to the instance during expansion rather than by name.

## The test

`boundOnModuleTypeParameterIsRejected` pins the rejection, in the same shape as `dispatchInsideConstructorIsRejectedForLua`, and its comment carries the scope rule above. Should the grammar change land, this test fails and becomes the success case.

Full suite green.

### DIFF
--- a/BACKLOG.md
+++ b/BACKLOG.md
@@ -77,8 +77,38 @@ itself, and one gap in what the suite can see.
 
     Do not start this autonomously.
 
-7. **Module bounds.** `module M<T: Show>` is rejected with a clear message today. Needs
-   receiver rewriting during expansion, or type parameters on `ModuleInstanciation`.
+7. **Module bounds.** `module M<T: Show>` is rejected with a clear message today, and
+   `TypeClassTests.boundOnModuleTypeParameterIsRejected` pins that. Tried and reverted; what follows
+   is why, because the earlier note here suggested a fix which cannot work.
+
+   Expansion copies the module body into the user and replaces the module's type parameters
+   **in type positions**. A requirement is called on the parameter itself — `T.show(x)` — and that
+   receiver is a name, resolved by `lookupBoundedTypeParam` through `lookupType`, so the replacement
+   never touches it.
+
+   Renaming the receiver to the using class's parameter, which is what "receiver rewriting during
+   expansion" meant, does not work. `NameResolution.nextScope` sends a `ModuleInstanciation` to
+   `attrModuleOrigin()` rather than to the class using it:
+
+       if (currentScope instanceof ModuleInstanciation) {
+           return nextScope(moduleInstanciation.attrModuleOrigin());
+       }
+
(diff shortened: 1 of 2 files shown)


Frotty:
Updated from master — the conflict was in `TypeClassTests`, where #1244 added its own bounded-parameter constant and tests at the same place as this branch's. Both are kept; `TypeClassTests` is green.

To be clear about what this pull request is, since it was asked whether it could implement the fix rather than record the finding: implementing it is a grammar change, and this branch does not attempt it. Receiver rewriting — the approach the backlog proposed — was tried and is ruled out, because `NameResolution.nextScope` sends a `ModuleInstanciation` to `attrModuleOrigin()` rather than to the class using it, so a receiver renamed to the using class's parameter names something that scope deliberately cannot see. The rename works; the name it produces cannot resolve.

What is left needs `ModuleInstanciation` to carry type parameters in `wurstscript.parseq`, so the copied body keeps saying `T` and `T` resolves against the instantiation. That means regenerating the AST and touching everything which reads that node, and it changes how expansion substitutes types — the body would resolve the parameter rather than have it replaced. It is a different size of change from this branch, and I would rather propose it separately than half-land it here.

So this stays as the pinned rejection plus the ruled-out approach. Say the word and I will take the grammar change on its own branch.

@codex review

PR #1247, opened 2026-08-16
Stop composing a dispatch slot which names no method

Frotty:
Backlog only — no behaviour change. Item 15 said a specialised class carries one dead dispatch slot and that fixing it risks more than it saves. Both halves are now verified rather than assumed, and the second turned out to be more true than the note knew.

## It is dead

In the Lua for `FastHashMap<int, int>`:

    FastHashMap_specialized_integer__integer.FastHashMap_specialized_integer__integer_integer
        = FastHashMap_get_specialized

Assigned once, bound to the alphabetically first method, never called — the only method calls in that script are `:create(` and `:create1(`. The name is the owner's plus the segment after the last underscore of the method's, and for a specialised method that segment is the type argument, so every method of the class composes the same name.

## Suppressing it in one composer is not enough

I changed `addDirectAliases` to drop a composed name which more than one method of the same class produces, on the grounds that a name meaning "one of these, arbitrarily" is worse than a name meaning nothing. All Lua dispatch, FastHashMap and type class tests passed — **and the slot was still emitted.**

`LuaTranslator.collectDispatchSlotNames` composes it independently, from the cross product of the class names in the hierarchy and the semantic names of the method group. It is called per group of same-named methods, so it cannot see that a sibling group composes the same name without class-level knowledge it is not given.

So a real fix needs a shared notion of ambiguity across both composers. That is the coordinated change the note weighed against the benefit, and the benefit is still one unused table key per specialised class. Reverted.

## What a fix would need

Recorded in the item, together with the earlier attempt: using the declared name instead collapses overloads, which two existing tests exist to prevent. Both sources of a semantic name are wrong in opposite directions — the mangled trailing segment collides across the siblings of one specialisation, the declared name across overloads — so the name has to separate both, in both composers. The declared name together with the dispatch signature key would.

### DIFF
--- a/BACKLOG.md
+++ b/BACKLOG.md
@@ -107,23 +107,36 @@ itself, and one gap in what the suite can see.
     `simplifyClasses` nor `addMemberTypeArguments`, so the type variables are never lifted there and
     a super call has nothing to carry. Closing it is item 23.
 
-15. **One junk dispatch slot per specialised class.** `addDirectAliases` and
-    `LuaTranslator.collectDispatchSlotNames` both compose `owner.getName() + "_" +
-    semanticNameFromMethodName(name)`, and for a specialised method that trailing segment is the
-    type argument — so every method of `FastHashMap<int, int>` claims one shared
-    `FastHashMap_specialized_integer__integer` slot and the alphabetically first wins it.
-    Nothing calls it, so it is dead weight rather than a wrong result.
+15. **One junk dispatch slot per specialised class.** Confirmed dead weight, and confirmed harder to
+    remove than it looks. Leave it unless it stops being dead.
 
-    Tried using the declared name instead and reverted it: overloads share a declared name, so
-    `setup(int)` and `setup(string)` collapse into one slot, which is what
+    In the Lua for `FastHashMap<int, int>` the slot
+    `FastHashMap_specialized_integer__integer.FastHashMap_specialized_integer__integer_integer` is
+    assigned once, bound to `FastHashMap_get_specialized` — the alphabetically first method — and
+    never called: the only `:` calls in that script are `:create(` and `:create1(`. The name is the
+    owner's plus the segment after the last underscore of the method's, and for a specialised method
+    that segment is the type argument, so every method of the class composes the same one.
+
+    **Two composers emit it, and suppressing it in one is not enough.** `addDirectAliases` was made
+    to drop a composed name which more than one method of the class produces — reasoning that a name
+    meaning "one of these, arbitrarily" is worse than a name meaning nothing — and the slot was still
+    emitted, because `LuaTranslator.collectDispatchSlotNames` composes it independently, from the
+    cross product of the class names in the hierarchy and the semantic names of the group. That
+    method is called per group of same-named methods, so it cannot see that a sibling group composes
+    the same name without being given class-level knowledge it does not currently have.
+
+    So a fix means a shared notion of ambiguity across both composers, which is the coordinated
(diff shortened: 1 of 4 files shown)


Frotty:
This now fixes it rather than recording it — the pull request title and scope have changed accordingly.

**The junk slot is gone.** Every slot on `FastHashMap<int, int>`'s specialised table now carries a method name, and the shared `FastHashMap_specialized_integer__integer_integer` key, which used to be bound to whichever method sorted first and never called, is no longer emitted.

**What made this harder than one composer.** Suppressing it in `addDirectAliases` alone changed nothing, because `LuaTranslator.collectDispatchSlotNames` composes the same name independently. Both now skip it.

**And the condition took two wrong attempts, which the tests caught.** My first was "more than one method of the class composes this name" — that breaks `genericOverrideChainBindsRootSlotToMostSpecificImplInLua`, and the failure is the point: in an override chain several methods *do* share a semantic name and *must* share a slot, because that is what dispatch is. My second used the alias sets as identity, which also fails, since overrides carry class-specific aliases and so differ.

What actually separates the two cases is the name in the source. A method and its overrides all declare the same one; the siblings of one specialisation declare different ones (`get`, `put`, `has`, `slotFor`) and merely end up composing the same segment, because for them that segment is the type argument rather than a method name. So the test is how many distinct declared names produce the composed name, and `declaredName` is now shared between the two composers instead of being private to one.

`everySlotOnASpecialisedClassNamesAMethod` pins the result by walking every assignment to the specialised table and requiring each slot to contain one of the container's method names.

Narrow tests green: `FastHashMapTests`, `LuaTranslationTests`, `TypeClassTests`.

@codex review

Frotty:
All three addressed. Two are fixed; the first is answered with a limit rather than a fix, and I want to be exact about why.

**Distinguish overloads when detecting ambiguous aliases.** You are right that `declaredName` misses them, and the FastHashMap test cannot see it because its methods have distinct declared names.

I tried the obvious sharper identity — the dispatch group key, which `assignDispatchGroupKeys` already computes by unioning a method with its submethods and splitting by signature. It fails, and the reason is worth recording: the key embeds the signature, and a generic override chain's signatures differ by each class's own type variable. Instrumented on the override repro:

    class=FindBuilder semantic=update keys=[
      BuildAtTarget|BuildAtTarget_update|void|Owner,real,
      NoOpState|NoOpState_update|void|T192,real,
      State|State_update|void|T636,real]

Three keys for one override chain, so the overrides read as unrelated and lose the slot they must share — `genericOverrideChainBindsRootSlotToMostSpecificImplInLua` fails exactly that way. I tried the alias sets as identity first, which fails for the same reason.

So the declared name is what survives, and the residual is precise: **a specialised class holding only overloads of one source method keeps one dead key**, because they share a declared name. It is dead weight as before rather than a new problem, but the change no longer claims the general case. Both composers say this where the decision is made, and backlog item 15 now describes it as what is left, with what a real identity would need — differing type variables treated as one signature while real parameter differences still separate.

**Cache ambiguity before iterating dispatch groups.** Fixed: computed once per receiver class and memoised, so `createMethods` asking twice per group no longer rebuilds and sorts the inherited method list each time.

**Mark the backlog item as completed.** Fixed. It said to leave the slot alone and that a fix had been reverted, which was the opposite of what this branch does.

Also merged master, since the backlog on this branch predated #1243 through #1246 and would have regressed their entries.

Narrow tests green: `FastHashMapTests`, `LuaTranslationTests`, `TypeClassTests`, `DeterminismDiffReportTest`.

@codex review

PR #1249, opened 2026-08-17
Let the interpreter ask what a node was copied from, and who owns it

Frotty:
Stage two, complete on this pull request. Consumes the relation from #1248, now merged, so this targets master directly.

Two name-based answers in `ProgramState`, both replaced by recorded ones. This closes the remaining half of backlog item 10.

## Type variables were compared by name

    boolean sameVar = e.getKey() == typeVar || e.getKey().getName().equals(typeVar.getName());

That takes two parameters which merely share a name for the same parameter — the mistake `EliminateGenerics` made, where it dispatched a string through an `int` instance and killed the interpreter (#1244). The same bug was waiting in the compiletime path, unpinned.

Its survival was legitimate rather than neglect: the interpreter is handed a program, not the translation which produced it, so it had nothing else to compare.

## A static field's owner was read out of its name

`identifyGenericStaticGlobals` took the **longest prefix of the global's name ending at an underscore which names a generic class**. A class whose name contains an underscore answers that wrongly, and silently — nothing downstream can tell a mis-attributed owner from a correct one.

The owner is now recorded where the specialised global is created, which is where it is known.

## How both reach the interpreter

`SpecialisationLookup` — two questions, `canonical(node)` and `genericStaticOwnerOf(global)` — handed over instead of the translator, keeping the interpreter's independence from the translation. `ImTranslator` implements it; `CompiletimeFunctionRunner` supplies the translator it already holds.

Without one, `SpecialisationLookup.NONE` makes every node its own original and the name search stands in. That is what a hand-built program in a test means, and it still answers those correctly.

One ordering detail worth naming: the owners are worked out in the `ProgramState` constructor, before the lookup is supplied, so supplying it re-derives them. I introduced that bug in this change and caught it before pushing — without the re-derivation the recorded path would never have fired, and the tests would still have passed, since the name search returns the same answer for every program in the suite.

## Behaviour

Unchanged where the two approaches agreed, which is everywhere the relation exists — that is why no test moves. What changes is that they can no longer agree by accident.

Green: `CompiletimeTests`, `TypeClassTests`, `FastHashMapTests`, `GenericsTests`, `GenericsWithTypeclassesTests`, `LuaTranslationTests`, `StdLibOwnTests` (all 460 library tests), `SpecialisationOriginTest`.

## What I did not fold in, and why

The naming half — `semanticNameFromMethodName` and the two slot composers. When I proposed stage two I said the relation would let names be composed from structure; #1247 showed that is wrong, and I would rather say so than quietly leave it out. An override chain's methods have **different** originals, so identity by origin splits a chain which must share one slot — which is exactly how the declared-name condition in #1247 was arrived at. The relation answers liveness and the identity of copies; it does not answer "is this method the same dispatchable thing as that one", and grouping a virtual method with its overrides needs that instead.

So the name parsing in the slot composers stays until there is an identity which handles override chains. Backlog item 15 records what such an identity has to do.

### DIFF
--- a/de.peeeq.wurstscript/src/main/java/de/peeeq/wurstio/CompiletimeFunctionRunner.java
+++ b/de.peeeq.wurstscript/src/main/java/de/peeeq/wurstio/CompiletimeFunctionRunner.java
@@ -90,6 +90,9 @@ public class CompiletimeFunctionRunner implements AutoCloseable {
         this.translator = tr;
         this.imProg = imProg;
         globalState = new ProgramStateIO(mapFile, mpqEditor, gui, imProg, true);
+        // The interpreter is handed a program; this hands over the one thing it cannot work out from
+        // the program alone, which is what a specialised node was copied from.
+        globalState.setSpecialisations(tr);
         initializeBackendConstants();
         this.interpreter = new ILInterpreter(imProg, gui, mapFile, globalState);
 
--- a/de.peeeq.wurstscript/src/main/java/de/peeeq/wurstscript/intermediatelang/interpreter/ProgramState.java
+++ b/de.peeeq.wurstscript/src/main/java/de/peeeq/wurstscript/intermediatelang/interpreter/ProgramState.java
@@ -11,6 +11,7 @@ import de.peeeq.wurstscript.intermediatelang.*;
 import de.peeeq.wurstscript.jassIm.*;
 import de.peeeq.wurstscript.parser.WPos;
 import de.peeeq.wurstscript.translation.imtojass.ImAttrType;
+import de.peeeq.wurstscript.translation.imtranslation.SpecialisationLookup;
 import de.peeeq.wurstscript.utils.LineOffsets;
 import de.peeeq.wurstscript.utils.Utils;
 import it.unimi.dsi.fastutil.ints.Int2ObjectOpenHashMap;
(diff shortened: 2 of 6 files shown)


Frotty:
@codex review

Frotty:
Right, and the objection was exact: the change altered which binding gets selected and no test moved, so nothing demonstrated either half.

`aBindingIsFoundByOriginRatherThanByName` builds what you described. A type argument frame holding two parameters both named `T` with different bindings, and a copy recorded as coming from only one of them. The unrelated parameter is inserted first, so a name comparison reaches it before the right one.

With the name comparison restored:

    AssertionError: the binding of the parameter this copy came from, not of one which shares its name
      expected [integer{show=Right(original_show)}] but found [integer{show=Right(unrelated_show)}]

The wrong instance — the same shape that killed the interpreter in `EliminateGenerics` (#1244). With the relation it returns the binding of the parameter the copy was made from. I ran it both ways to confirm rather than asserting it would fail.

It also covers the wiring implicitly: the test supplies the translator through `setSpecialisations`, and without that supply the lookup is `NONE`, every node is its own original, and the copy finds no binding at all.

One thing your comment asked for that this does **not** cover: the owning-class half, `identifyGenericStaticGlobals`. The name search and the recorded answer agree for every program in the suite, and a program where they disagree needs a class whose name contains an underscore in a position that makes the longest-prefix search pick the wrong class. I did not manage to build one that reaches the interpreter, so that half rests on the recorded answer being known-correct at the point it is recorded rather than on a failing case — and I would rather say so than imply the new test covers it.

Green: `SpecialisationOriginTest`, `CompiletimeTests`, `TypeClassTests`.

@codex review

PR #1250, opened 2026-08-17
Decide slot ambiguity by the dispatch family, not by a declared name

Frotty:
The naming half of the root-level work — and a correction to something I said while proposing it.

## What I got wrong before

Reviewing #1247 I concluded there was no identity that separates overloads while keeping an override chain together, and settled for the declared name. I also told the owner the specialisation relation from #1248 could not help here. Both conclusions were premature.

`assignDispatchGroupKeys` already computes the right thing. It unions a method with its submethods — a method together with its overrides, which is one dispatchable thing. What it *stores* is taken after that union is split by signature, and the split separates the members of a generic override chain, because their signatures differ by each class's own type variable:

    class=FindBuilder semantic=update keys=[
      BuildAtTarget|BuildAtTarget_update|void|Owner,real,
      NoOpState|NoOpState_update|void|T192,real,
      State|State_update|void|T636,real]

I had measured that and drawn the wrong conclusion from it: the problem is the *split*, not the union. The union root, recorded before the split, is the identity that was wanted.

## The change

`ImMethod` gains `luaDispatchFamilyKey`, set from the union root. Both slot composers decide ambiguity from it rather than from the declared name.

- overrides share a family, so the slot they must share survives
- overloads and unrelated siblings do not, which the declared name cannot express

## What this does not demonstrate

#1247 recorded a residual: overloads inside a specialised class keeping a dead key, because they share a declared name. I wrote a test for it — two overloads of `get` in the specialised container — and **it passes under the declared name as well**, because the two overloads compose different names rather than colliding.

So the residual is narrower than that item claimed, or not reachable that way. This is a cleanup that removes the gap in principle, not a fix with a failing case behind it, and I would rather say that than let the commit imply otherwise. The test is kept for the shape, labelled with what it does and does not show, and item 15 is corrected rather than closed.

The change is still worth landing on its own terms: the family key says what the code means, where the declared name was standing in for it.

Green: `FastHashMapTests`, `LuaTranslationTests`, `TypeClassTests`.

## Where the root-level work now stands

- Stage 1, the relation — merged (#1248)
- Stage 2, the interpreter reading it — merged (#1249)
- Stage 2, naming — this pull request
- Stage 3, module expansion as an environment — parked on `feat/module-instanciation-type-params`, blocked on a decision: generic modules already resolve their type parameters by matching the receiver type, and declaring the parameters on the instantiation collides with that
- Stage 4, the Lua erasure model — the owner's decision, backlog item 23

### DIFF
--- a/de.peeeq.wurstscript/parserspec/jass_im.parseq
+++ b/de.peeeq.wurstscript/parserspec/jass_im.parseq
@@ -70,6 +70,11 @@ ImMethod(@ignoreForEquality de.peeeq.wurstscript.ast.Element trace,
 	java.util.List<ImMethod> subMethods,
 	java.util.List<String> luaMethodDispatchAliases,
 	String luaDispatchGroupKey,
+	// A method and its overrides, before the group is split by signature. Overrides of one method
+	// share this; overloads and unrelated methods do not. The group key cannot answer that, because
+	// it embeds the signature and a generic override chain's signatures differ by each class's own
+	// type variable.
+	String luaDispatchFamilyKey,
 	boolean isAbstract)
 	
 	
--- a/de.peeeq.wurstscript/parserspec/jass_im.parseq
+++ b/de.peeeq.wurstscript/parserspec/jass_im.parseq
@@ -70,11 +70,6 @@ ImMethod(@ignoreForEquality de.peeeq.wurstscript.ast.Element trace,
 	java.util.List<ImMethod> subMethods,
 	java.util.List<String> luaMethodDispatchAliases,
 	String luaDispatchGroupKey,
-	// A method and its overrides, before the group is split by signature. Overrides of one method
-	// share this; overloads and unrelated methods do not. The group key cannot answer that, because
-	// it embeds the signature and a generic override chain's signatures differ by each class's own
-	// type variable.
-	String luaDispatchFamilyKey,
 	boolean isAbstract)
 	
 	
--- a/de.peeeq.wurstscript/src/main/java/de/peeeq/wurstscript/translation/imtranslation/ClosureTranslator.java
+++ b/de.peeeq.wurstscript/src/main/java/de/peeeq/wurstscript/translation/imtranslation/ClosureTranslator.java
@@ -195,7 +195,7 @@ public class ClosureTranslator {
         tr.getImProg().getFunctions().remove(impl);
         c.getFunctions().add(impl);
         ImClassType methodClass = JassIm.ImClassType(c, JassIm.ImTypeArguments());
-        ImMethod m = JassIm.ImMethod(e, methodClass, superMethod.getName(), impl, JassIm.ImMethods(), Lists.newArrayList(), "", false);
+        ImMethod m = JassIm.ImMethod(e, methodClass, superMethod.getName(), impl, JassIm.ImMethods(), Lists.newArrayList(), "", "", false);
         c.getMethods().add(m);
 
         OverrideUtils.addOverrideClosure(tr, superMethod, m, e);
--- a/de.peeeq.wurstscript/src/main/java/de/peeeq/wurstscript/translation/imtranslation/ClosureTranslator.java
+++ b/de.peeeq.wurstscript/src/main/java/de/peeeq/wurstscript/translation/imtranslation/ClosureTranslator.java
@@ -195,7 +195,7 @@ public class ClosureTranslator {
         tr.getImProg().getFunctions().remove(impl);
         c.getFunctions().add(impl);
         ImClassType methodClass = JassIm.ImClassType(c, JassIm.ImTypeArguments());
-        ImMethod m = JassIm.ImMethod(e, methodClass, superMethod.getName(), impl, JassIm.ImMethods(), Lists.newArrayList(), "", "", false);
+        ImMethod m = JassIm.ImMethod(e, methodClass, superMethod.getName(), impl, JassIm.ImMethods(), Lists.newArrayList(), "", false);
         c.getMethods().add(m);
 
         OverrideUtils.addOverrideClosure(tr, superMethod, m, e);
--- a/de.peeeq.wurstscript/src/main/java/de/peeeq/wurstscript/translation/imtranslation/ImTranslator.java
+++ b/de.peeeq.wurstscript/src/main/java/de/peeeq/wurstscript/translation/imtranslation/ImTranslator.java
@@ -889,7 +889,7 @@ private void callInitFunc(Set<WPackage> calledInitializers, WPackage p, @Nullabl
         public ImMethod initFor(StructureDef classDef) {
             ImFunction impl = destroyFunc.getFor(classDef);
             ImMethod m = JassIm.ImMethod(classDef, selfType(classDef), "destroy" + classDef.getName(),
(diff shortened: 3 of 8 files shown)


Frotty:
Correct, and it broke the approach rather than needing a patch on top of it — so the family key is withdrawn and replaced with something that does not need families at all.

**Your reading was right.** `OverrideUtils.addOverride` links `superMethodIm -> wrapperMethod` and never `wrapperMethod -> subMethod`, so the union put the wrapper and the real override in different families and their shared slot was suppressed as a collision.

**The suggested fix is not safe.** I added `wrapperMethod.getSubMethods().add(subMethod)` and it fails six tests — `implicitConversionFail`, `implicitConversionFail2`, `implicitConversions3`, `genericFunctionOverload` and two closure-event Lua tests. The wrapper's dispatch then reaches the override directly, skipping the index conversion the wrapper exists to perform. The relation you describe is real, but stating it as a submethod link makes the wrapper redundant. I have left a note where the wrapper is built so the next reader does not repeat it.

**What replaces it.** A composed name is junk when it **names no method**. For a specialised method the trailing segment is the type argument, which nobody declared, so every method of the specialisation composes it and the slot is claimed by whichever is bound first. A method and its overrides compose the name they were *declared* with and must share that slot — and a conversion wrapper carries that same declared name, so it falls on the right side without any linking.

The `luaDispatchFamilyKey` field is gone from the grammar with it, rather than left unused.

It also covers the case item 15 recorded and the declared name alone could not: two overloads inside a specialised class both compose the type argument, which is neither of their declared names, so the slot is left uncomposed.

Green: `GenericsTests`, `LuaTranslationTests`, `FastHashMapTests`, `TypeClassTests`, `GenericsWithTypeclassesTests`.

@codex review

Frotty:
Fixed, and thank you for catching it — this was a nondeterminism I introduced in the change meant to make naming structural, which is the opposite of the point.

Two distinct methods can share a 32-bit `System.identityHashCode`, and the value depends on where the JVM allocated them. Using it as a claimant key made two methods read as one, so the slot could appear or disappear between runs for identical input. `luaOutputIsDeterministicForGenericOverrideSlots` exists to catch exactly that, and it would have been a coin toss whether it did.

Claimants are counted by object identity now, through a set backed by an `IdentityHashMap`, in both places. The decision is also stated outright rather than smuggled into the key format:

    a name is junk when nobody declared it and more than one method composes it

Two maps per owner — whether any method declared the name, and which methods composed it — instead of one map of encoded strings. It reads as the rule it is, and there is nothing left in it that depends on allocation.

Green: `FastHashMapTests`, `LuaTranslationTests`, `GenericsTests`, `TypeClassTests`.

@codex review

Frotty:
Right, and it invalidated the shape of the check rather than a detail of it, so the per-name decision is gone.

One method declaring a name should not entitle the rest of the specialisation to the slot it owns, which is exactly what `declaredByAnyone` did. The test is per method now:

> a method composes the class-qualified name only when the segment it would be built from is **its own** declared name

A specialised method's segment is the type argument, which names no method, so it composes nothing regardless of what any sibling is called. Overrides carry the declared name and keep the slot they share, and a conversion wrapper carries it too — so that case is covered without linking, as before.

This also removes the machinery rather than fixing it: no buckets, no identity sets, no counting, and no `ambiguousSemanticNames` cache. The decision is local to one method, which is why nothing another method is called can sway it. That is a better answer than the two I had before it, and it came out of your example.

**I could not build your exact scenario as a test.** `function integer()` is rejected — *"The name 'integer' is a built-in type and cannot be used here"* — and for a class type argument the name would clash with the class. I wrote the test, hit that, and removed it rather than leave a broken one. So the fix rests on being strictly narrower than what it replaces: it composes a name in a subset of the cases the previous rule did, and the case you describe is one of the ones it now refuses.

Green: `FastHashMapTests`, `LuaTranslationTests`, `TypeClassTests`, `GenericsTests`, `GenericsWithTypeclassesTests`.

@codex review

Frotty:
Confirmed and fixed in de578a65f. It is a regression this PR introduced, not something it inherited.

I wrote the case as a test before touching anything - a generic base and a subclass each declaring two overloads of one method, called through a base-typed reference - and it fails on this branch. Then I checked out the previous composers on the same branch and ran it again: it passes. So the rule here is what breaks it, exactly as described.

The mechanism is the one you named. A method the translation numbered to tell it from its overloads carries that number in the segment the slot name is composed from, so the second `route` mangles to `Base_route1` and the segment is `route1` while the declaration says `route`. Requiring equality refuses the alias for every overload past the first, the override never composes the class-qualified slot, and the call stays bound to the base.

The fix allows the segment to be the declared name followed by that number, since the number is the translation's own and means the same method. Nothing else is allowed between them, which keeps the case the check exists for: a segment which is a type argument rather than a method, composing a slot that names nothing.

Worth recording that this is the second thing the same equality has refused. Auditing it earlier turned up a declared name containing an underscore - `get_it` yields segment `it` - which breaks dispatch in a generic hierarchy on master too, so that one is pinned as a known failure on #1251 rather than fixed here. Both come from recovering the declared name by cutting the mangled one at its last underscore, and the durable fix is keyed on the declared signature. Backlog item 26 carries it, including why pointing both composers at `declaredName` alone does not work: two overloads share a declared name, so that collapses their slots instead.

Green: `LuaTranslationTests`, `FastHashMapTests`.

@codex review

Frotty:
Checked this one rather than reasoned about it, and it does not reproduce.

I built the case described - a class named `route1` used as the type argument of a generic which declares both `route` and `route1`, same arity, called through the specialisation - and both calls bind to their own implementations. Kept as a test (`aTypeNamedLikeAnOverloadNumberDoesNotStealTheSlot`) rather than left as an argument.

The reason is structural, not luck: the type argument becomes part of the owning class's name, not the tail of the method's. The segment a slot name is composed from is `route` for `route` and `route1` for `route1`, so both match their declared names exactly and the numeric tolerance is never reached. For the tolerance to be reachable at all, the type argument would have to be the last segment of the method's own mangled name, which is a different shape from the one described here.

That said, the underlying point stands and I am not going to keep patching around it. This predicate has now produced three findings: a declared name containing an underscore (real, fails on master too, pinned on #1251), an overload number (real, a regression this PR introduced, fixed in de578a65f), and this one (not reachable). Two out of three were genuine, and all three exist because the declared name is being recovered by cutting a mangled name at its last underscore instead of being asked for. Overload identity determined structurally is the right answer and it is backlog item 26, which also records why the obvious version of it does not work - pointing both composers at `declaredName` collapses overloads, which two existing tests catch.

So: this PR keeps the narrow tolerance with the reachable case fixed and covered, and item 26 replaces the parse with a signature key rather than adding a fourth exception to it.

Green: `LuaTranslationTests`, `FastHashMapTests`.

Frotty:
@codex review

Frotty:
Closing into #1253. The rule here was a string heuristic getting exceptions bolted onto it - a declared-name equality, then a tolerance for the number the translation appends to overloads - and the review comments were finding the next hole each time because the shape was wrong, not the details.

Master already refuses to compose a slot naming no method (dea459b45), so nothing here is lost by closing it. The two tests worth keeping moved to #1253 and pass against unmodified master: an override of a numbered overload reached through its base, and a type argument named like an overload number not stealing a slot.

PR #1251, opened 2026-08-17
Record the four decisions, and settle div and mod

Frotty:
The owner answered the four open questions. This writes them into the backlog so they survive the session, and closes the one that needed no code.

## Settled: `div` and `mod` keep returning the left operand's type

So `real r = 7 div 2` compiles, and it is meant to. Item 8 has been pending a yes or no since early in this work, with the behaviour pinned in both directions meanwhile.

The branch in `AttrExprType` now says so, rather than looking like an oversight beside `caseMathOperation` — which collapses two integer literals to int precisely so `real r = 1 + 1` is an error. The asymmetry is deliberate: the division is integer either way and the result is then widened.

`ExpressionTests.integerDivisionOfLiteralsIsStillAssignableToReal` and `OptimizerTests.realFormatting_consistent_fromIntOps` both still pass, which is the point — the second opens with that assignment.

## Recorded: module bounds (item 7)

The instantiation declares the module's type parameters **only so a dispatch receiver has a name to resolve**, and they stay out of type inference. Generic modules keep resolving theirs by matching the receiver type.

The alternative — letting inference see the declared parameters — collides with that mechanism and fails `genericModuleInGenericClassGet` with *"Cannot infer type for type parameter T"*. This is the smaller change, at the honest price of the parameter meaning something narrower than it appears to.

The item now carries what the attempt reached, so it is resumable rather than rediscovered: the grammar change, the binding through `WurstTypeBoundTypeParam`, the dispatch widening, and the one step left — the requirement lookup does not follow a binding to the underlying parameter's bounds.

## Recorded: the Lua erasure model (item 23)

Specialise only the paths which need a concrete type and leave the object erased. Generated scripts stay small, which is the reason; the cost is that it is more compiler work than not erasing at all.

The item spells out what that means for items 6 and 13 — the concrete type is threaded to the places which use it rather than to the object — and why it is worth doing rather than working around: two class shapes existing at once is what produced both field bugs in #1239.

## Deferred, with the reason: instances for a family of types

It decides whether type class bounds stay a tool for new containers or become how `Table`, `HashMap`, `HashList` and `HashSet` work. Worth deciding when there is appetite for the language design, rather than alongside compiler work.

Green: `ExpressionTests`, `OptimizerTests`.

### DIFF
--- a/BACKLOG.md
+++ b/BACKLOG.md
@@ -189,6 +189,34 @@ itself, and one gap in what the suite can see.
     path, and `transformGenericNewOnly` does not lift them, so the method's own parameter is counted
     against a list which does not include the class's.
 
+26. **Nothing in the compiler should recover structure from a name, and one place doing it is a bug
+    today.** A dispatch slot's name is composed from the segment after the last underscore of a
+    method's mangled name, which is the declared name only when the declared name has no underscore in
+    it: `get_it` contributes `it`, which is nobody's method.
+
+    That is not only untidy. `LuaTranslationTests.underscoreNamedOverrideInAGenericHierarchyIsStillBrokenOnLua`
+    pins the consequence - an override named `get_it` in a generic hierarchy does not dispatch on Lua,
+    while the same shape without the underscore does. Found by auditing the junk-slot rule for what its
+    name comparison does to unrelated methods, not by anyone hitting it.
+
+    Asking the declaration instead is not the whole fix, and trying it is how the rest of this was
+    found. `LuaDispatchPreparation.declaredName` already reads the name off the trace, but two
+    overloads share a declared name: they mangle to `Foo_bar` and `Foo_bar_1`, and the segment after
+    the last underscore is also what currently keeps their slots apart. Pointing both composers at the
+    declared name fixes the underscore case and collapses overloaded slots instead -
+    `overloadedMethodsDoNotAliasInLuaDispatchTables` and
+    `moduleProvidedOverloadedOverrideDoesNotCollapseLuaSlots` both catch it.
+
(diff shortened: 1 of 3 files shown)


Frotty:
Added the underscore dispatch finding here, since it came out of auditing the junk-slot rule rather than out of anything on this branch.

A dispatch slot's name is composed from the segment after the last underscore of a method's mangled name. That is the declared name only when the declared name has no underscore in it, so `get_it` contributes `it` - nobody's method - and an override named `get_it` in a generic hierarchy does not dispatch on Lua. The same shape without the underscore works, which is why nobody has hit it. Present on master, pinned as a known failure with what it currently produces.

I tried the obvious fix and it does not hold, which is the more useful half. `LuaDispatchPreparation.declaredName` already reads the name off the trace, so pointing both composers at it looked like a rename. But two overloads share a declared name and mangle to `Foo_bar` and `Foo_bar_1`, and the segment after the last underscore is also what currently keeps their slots apart - so that change fixes the underscore case and collapses overloaded slots instead. `overloadedMethodsDoNotAliasInLuaDispatchTables` and `moduleProvidedOverloadedOverrideDoesNotCollapseLuaSlots` both catch it.

What a method and its overrides share, and what tells two overloads apart, are two different questions, and the mangled name is answering both at once by accident. The replacement is keyed on the declared signature, which is a real change rather than a rename, so it is backlog item 26 with that reasoning written down rather than something bundled in here.

@codex review

Frotty:
Closing into #1253, which carries the pinned dispatch failure and the backlog entry with the evidence from both failed attempts at the fix.

The div/mod decision and the other recorded decisions are in that branch's BACKLOG.md unchanged.

PR #1253, opened 2026-08-17
Make a dispatch slot name arrive as data instead of being recovered from another name

Frotty:
One PR for the whole naming problem. #1250 and #1251 are closed into this - the tolerances they were accumulating are gone, and only what encodes real behaviour is kept.

## The problem

A Lua dispatch slot's name is composed by cutting a method's name at its last underscore and taking the tail. That tail is the declared name only when the declared name has no underscore in it and the method is not a specialised copy. Every way that assumption fails has produced a bug:

| Shape | Consequence |
|---|---|
| `get_it` | tail is `it`, nobody's method; an override in a generic hierarchy does not dispatch |
| specialised method | tail is the type argument; slot named after a type, not a method |
| numbered overload | tail is `route1` where the declaration says `route`; an override of it cannot replace its ancestor's slot |
| type named like an overload number | could collide with a tolerance for that number (not reachable) |

## What is in this commit

Three tests, one per shape that matters, all against unmodified master. The underscore case is asserted as the failure it currently produces, so it is pinned rather than met by surprise later. The other two pass and stay as regression cover.

## Why the obvious fixes are not it

Both were tried, and both failures point at the same conclusion.

Asking `declaredName` alone collapses overloads - two overloads share a declared name, and `overloadedMethodsDoNotAliasInLuaDispatchTables` catches it. Using the method's name whole rather than its tail breaks cross-class matching, because at the point slots are composed a method's name is still class-prefixed:

```
GlobalCheckState.State_update = GlobalCheckState_GlobalCheckState_update
```

The ancestor's slot is `State_update` while the method is `GlobalCheckState_update`. The cut is load-bearing precisely because the prefix is there, which is why every attempt to keep the cut and qualify it has needed another exception.

## What lands on this branch next

`ImMethod` carries its declared name and the index the translation gave it among its overloads, recorded where the translation assigns them, and slots compose from that pair. `luaDispatchGroupKey` is already a recorded field on the method, so the shape exists. Then the cut is deleted rather than tolerated, both composers ask one question, the pin above starts passing, and `ProgramState.identifyGenericStaticGlobals` - which takes the longest prefix of a global's name ending at an underscore that matches a class name - gets the same treatment.

Green: `LuaTranslationTests` (the pin failing as asserted).

### DIFF
--- a/BACKLOG.md
+++ b/BACKLOG.md
@@ -31,6 +31,12 @@ itself, and one gap in what the suite can see.
    a repro for both targets, `TypeClassTests.dispatchInsideConstructor` and
    `dispatchInsideConstructorIsRejectedForLua`, the second pinning the current diagnostic.
 
+   Still open after item 23's first half landed. The two mechanisms that closed items 13 and 25 both
+   read the instantiation off a receiver — the first argument of a call which names its target, or the
+   receiver of a method call. A constructor has neither: `new_Box(21)` takes no receiver, and the only
+   thing stating `Box<integer>` is the type of the variable it is assigned to. So this still needs the
+   type-driven collector below, which is a second mechanism rather than a widening of the one there is.
+
    Not the same gap as item 5, and the fix from it does not reach: a constructor belongs to the
    class rather than to a generic function of its own, so the call that runs it carries no type
    arguments at all. The intermediate language has `b = new_Box(21)` with `b` typed
@@ -62,7 +68,22 @@ itself, and one gap in what the suite can see.
     instance be allocated with no fields at all, and then with its fields under a key nothing read
     (#1239). Both are fixed; the shape which produced them is what this decision removes.
 
-    Unblocks items 6 and 13's Lua half. Not blocking the container, which works on both targets today.
+    **Half done.** Items 13 and 25 are closed, and the two things that closed them are the model
+    working: a specialised method is moved to the class its objects are actually allocated from, and a
+    call which names its target reads the instantiation off its receiver argument rather than off the
+    object's class. A specialisation nothing allocates is now left with no methods and drops out
+    entirely, so the second class shape is gone wherever an ordinary generic object is involved -
+    `FastHashMapTests.assertSpecialisedClassesAllocateTheirFields` states both accepted outcomes rather
+    than the one that used to hold.
+
+    What is left is item 6, and it is the part this decision said would be the work: reading the
+    instantiation from a *type* rather than from a receiver. The classes which still allocate a
+    specialised copy are the ones whose construction was itself specialised, which is the same
+    question. Deciding it from what the program allocates, as the pass above does, is a fact rather
+    than a guess - but it is decided after the fact, and the model would rather no specialised class
+    were ever allocated.
+
+    Not blocking the container, which works on both targets today.
 
 9. **Keep `WURST_LANGUAGE.md` and `CHANGELOG.md` current** as items land — a standing practice
    rather than a task to finish. `WURST_LANGUAGE.md` is tracked, at
@@ -150,22 +171,6 @@ itself, and one gap in what the suite can see.
     far produced nothing to work from, which is why it cost a re-run and no diagnosis. The next one
     will say what differed.
 
-25. **A bounded type parameter on a method of a generic class is rejected on Lua.**
-    `class Holder<T: Show>` with `function convert<Q: Show>(Q other)` fails there with "Generics should
-    match class method type variables", while the same program compiles and runs on the other target.
-    On master as well, so it is not a regression — found while trying to give item 10's fix Lua
-    coverage, which is what it blocks: the only shape reaching that lookup with two parameters at once
-    is a class parameter beside a method parameter, and Lua will not compile it.
-
-    Pinned by `TypeClassTests.aBoundedMethodParameterInAGenericClassIsRejectedForLua`. A version with
-    the second parameter on a free function does compile on Lua and passes with or without item 10's
-    change, so it covers nothing; that is why the rejection is pinned instead.
-
-    Where to start: the message comes from the arity check between a call's generics and the callee's
-    type variables. A method of a generic class has the class's variables lifted onto it on the Jass
-    path, and `transformGenericNewOnly` does not lift them, so the method's own parameter is counted
-    against a list which does not include the class's.
-
 26. **Done. A dispatch slot's segment is recorded where it is assigned, not recovered from a name.**
     The segment used to be found by cutting a method's name at its last underscore, which is the right
     answer only when the rest contains no underscore and the method is not a specialised copy. Four bugs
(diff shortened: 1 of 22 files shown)


Frotty:
On the pinned test: agreed that a pin is not the end state, and it is there as a placeholder for exactly the fix you are asking for, not as a claim that base-bound dispatch is correct. When the fix lands, TestNG failing for a missing expected exception is the signal to flip it to a positive assertion - which is the point of pinning it that way rather than leaving the case unwritten.

I have now tried the fix twice more on this branch and reverted both, and the failures narrow it down usefully.

Using the method's name whole instead of its tail breaks cross-class matching: `GlobalCheckState_update` where the ancestor's slot is `State_update`.

Stripping the owner's name as a known prefix - the boundary is not a guess, the owner is right there - does fix the underscore case: the pin flipped to passing, which is the first time anything has. But it broke `genericOverrideChainBindsRootSlotToMostSpecificImplInLua` and `genericOverrideChainBindsGlobalStateSlotToMostSpecificImplInLua`, and the reason is the part I had wrong. `normalizeMethodNames` assigns one name per dispatch group, derived from the first member's already class-prefixed name, and sets it on every member. So the prefix a method's name carries is not necessarily its own owner's - an ancestor's method can be named after a descendant's class. Cutting at the last underscore is robust to that by accident, which is why it has survived; no prefix known locally is the right anchor.

That settles the design rather than leaving it open. The segment has to be recorded at the one point where both the group and the name assigned to it are known, which is `normalizeMethodNames`, as a field on `ImMethod` beside `luaDispatchGroupKey` - so a grammar change and `genAst`, then both composers read it and the cut is deleted. The backlog entry on this branch has been updated with all three failed attempts so the next one starts past them.

The tests and that entry are what this PR carries today, all against unmodified master, with `LuaTranslationTests` green.

@codex review

Frotty:
Correct, and it is a stale paragraph rather than a disagreement - the commit before this one had already concluded the recording point is `normalizeMethodNames`, but the plan further up the entry still said "where the translation assigns them", which is the earlier point you are objecting to.

Fixed in f44a32d28, and the entry now states why the earlier point is not equivalent: the assigned name may be derived from a different member of the group, and sanitising it into a Lua identifier and uniquing it against everything already taken can change it. A declaration-derived pair would have to be matched back to whatever came out of that, which is the recovery problem again under a new name.

So the field is filled in `normalizeMethodNames`, where the group and the name assigned to it are both in hand, and both composers read the field rather than cutting a string.

@codex review

Frotty:
Right, and the claim was false rather than merely unhelpful - the entry said the branch explains itself and there was no comment there at all. Fixed by doing what the item tracked instead of reopening it: 6c0f8cd adds the rationale to the `MOD_INT`/`JASS_MOD_INT`/`DIV_INT` branch in `AttrExprType`.

It says what the decision was and why the neighbouring code looks like it contradicts it: returning the left operand's type is what makes `real r = 7 div 2` compile, while `caseMathOperation` a few lines down collapses two int literals to int so `real r = 1 + 1` is an error. These operators are integer-only, an int literal is a subtype of real, and narrowing the result would break assignments which compile today. Both tests that hold it are named in the comment, so someone changing the branch sees them without going through the backlog.

`ExpressionTests` green.

Frotty:
Correction to the comment above: the commit is 929235081, not 6c0f8cd. The change and the reasoning are as described.

Frotty:
@codex review

Frotty:
Implemented in 785115580 - the cut is gone and the pinned test is now a positive one, which answers the review comment above directly.

`normalizeMethodNames` names a dispatch group after one of its members, sanitises that name into a Lua identifier and uniques it against every name taken. It now strips the naming member's class from the result and records the segment on `ImTranslator` for every member of the group. Both composers read the record, and `semanticNameFromMethodName` is deleted from both.

Nothing else in the compiler can compute this, which is why the three earlier attempts failed:

- asking the declaration alone collapses overloads, since two overloads share a declared name
- using a method's name whole breaks cross-class matching, since the name is class-prefixed
- stripping each method's own owner fixes the underscore case and breaks override chains, because a group is named after one member and an ancestor's method can carry a descendant's class in its name

That last one is the reason the recording has to happen where the group is in hand rather than anywhere a method can reach on its own.

`underscoreNamedOverrideDispatchesInAGenericHierarchy` passes rather than being pinned as broken. Green: `LuaTranslationTests`, `FastHashMapTests`, `GenericsTests`, `ClosureTests`, `TypeClassTests`, `OptimizerTests`, `DeterministicChecks`, `StdLibOwnTests`.

@codex review

Frotty:
@codex review

Frotty:
@codex review

Frotty:
Implemented in b78e302ae, with one correction to the reasoning.

The condition did read the two shapes as alternatives, and that is not justified: a call is rewritten to the specialised `ImMethod` whichever class its receiver came from, so skipping the rebind purely because the specialised class is allocated leaves whichever shape lost the method without the slot. The erased class now gets a binding of its own to the same implementation and the specialised class keeps what it has. Being allocated is what decides where a slot is needed, on both sides.

Where the review is not quite right is the consequence. I could not produce a failing program, and I did look:

- The branch fires today only for a closure class, `Producer_test` against `Producer_test_specialized_integer`. There the erased class already carries a binding under the specialised slot name, and its allocation is dead in the emitted script, so the tables come out byte-identical before and after this commit.
- The non-closure route I expected to reach it does not. A generic function which constructs the class is specialised, but its copy still allocates the *erased* class, so only one shape is ever allocated and the existing move already applied.

So this is the rule stated rather than a defect I can show failing, which is why the commit says so.

The new test covers the other half of that question instead, since it is the part I could make fail: `oneGenericReachedThroughEveryConstructionLua` has three receivers of one class - built by a specialised generic function, constructed directly, and a subclass instance - all reaching one specialised slot that the override makes virtual. Because a specialised function copy allocates the erased class too, the move has to serve every way of arriving at the slot and not only the construction written in place.

Green: `LuaTranslationTests`, `FastHashMapTests`, `GenericsTests`, `GenericsModuleTests`, `ModuleTests`, `ClosureTests`, `TypeClassTests`, `OptimizerTests`, `DeterministicChecks`, `StdLibOwnTests`.

@codex review

Frotty:
@codex review

Frotty:
Both addressed in 34cf0aeff, one accepted and one reverted.

**Restrict receiver-based specialization — accepted.** This was a real miss. `collectCallThroughGenericReceiver` specialised whatever it could read an instantiation for, without asking whether the target reaches a dispatch or a construction, so every call into a generic superclass would get a copy per instantiation of functions that need none. The same `functionNeedsSpecialization` check the other collectors use now gates it. Being able to read an instantiation says nothing about whether anything wants it.

**Link the erased copy into the dispatch group — reverted, and the copy with it.** The reasoning here is right and it is what makes the suggestion from the previous round unworkable, so I have taken that one back out.

Following it through: an unlinked copy is a dispatch group of its own, is named separately, and lands under a name no call site asks for - so the binding was useless as written, as you say. Linking it with `method.getSubMethods().add(onErased)` then fails outright:

```
Cannot change parent of element ImMethodImpl, as it is already used in another tree.
```

and with that resolved it would merge two groups which are distinct on purpose. `dispatchInsideClosureLua`, `dispatchInsideClosureThroughHelperLua` and `nestedClosuresInsideBoundedGenericAreRejectedForLua` all fail on it.

So the branch is gone and the methods stay where they are. The premise behind adding it does not hold for the one shape that reaches it: a closure, where each class already binds its own implementation under the same slot names and the erased allocation is dead in the emitted script. I could not construct a case where an erased instance reaches a slot only the specialised class carries - two attempts, both ending with only one shape allocated. The reasoning is recorded at the branch so the next attempt starts from what happened rather than from how it looks.

The emitted closure script is byte-identical to before either change.

Green: `LuaTranslationTests`, `FastHashMapTests`, `GenericsTests`, `GenericsModuleTests`, `ModuleTests`, `ClosureTests`, `TypeClassTests`, `OptimizerTests`, `DeterministicChecks`, `StdLibOwnTests`.

@codex review

PR #1254, opened 2026-08-18
Fix the macOS-only library failure, and say which test it was

Frotty:
### The macOS failure, fixed

`macos-15-intel` was the only failing job on master, for several runs, while `ubuntu-latest`, `windows-latest` and `macos-latest` all passed. It was `StdLibOwnTests.standardLibraryTestsPass`, one library test of 460.

Two different causes were mixed into that history, worth separating:

- Run `32044793479` failed in 59s on `actions/setup-java` returning 503 then 429. GitHub infrastructure, not a test.
- Runs `32072039746` and `32028255014` are the real one: 459/460, on the slowest runner only.

The test was `PolygonTests.acceleratedClassificationMatchesLinearFor10000Points`. Both strides in its point sequence are coprime to 1152, so ten thousand iterations asked the same 1152 questions nearly nine times over — no extra coverage, and enough interpreted work to reach the twenty second per-test budget on a loaded machine. WurstStdlib2#467 fixed that; this bumps the pinned library from `98b1140` to `a85001e`, which carries it.

Verified against the new pin: **460/460**. The test is now named `acceleratedClassificationMatchesLinearForEveryBenchmarkPoint` and reports 1152 lookups rather than 10000; the count is unchanged because it was made cheaper rather than removed.

### And why placing it was so hard

A failing library test threw the counts followed by every warning the library compiles with. The test's name went only to stdout, which a CI run does not keep, so the failure said one test of 460 had failed and nothing else. Placing it took three runs, the check annotations and the published summary, and still came down to inference — the annotation was cut off at around 8000 characters of warnings before reaching anything useful.

`RunTests` already records each failure with its function and message; only the counts were being passed on. They are now named in the thrown message, ahead of the warnings, because a report which truncates a long message keeps the front of it.

The name is package qualified, reusing `RunTests.qualifiedTestName` rather than spelling it a second time: `standardLibraryTestsPass` imports every test package the library has, so two of them naming a test the same would otherwise still be indistinguishable.

### Testing

`StdLibOwnTests.aFailingLibraryTestIsNamed` runs a deliberately failing `@Test` and asserts the thrown message names it, qualified. It cannot pass vacuously: no error fails the test, and a message without the name fails the assertion.

Green: `StdLibOwnTests` (both tests) and `GenericsWithTypeclassesTests`, the other suite using this path.

### DIFF
--- a/de.peeeq.wurstscript/src/test/java/tests/wurstscript/tests/StdLib.java
+++ b/de.peeeq.wurstscript/src/test/java/tests/wurstscript/tests/StdLib.java
@@ -24,7 +24,7 @@ public class StdLib {
     /**
      * version to use for the tests
      */
-    private final static String version = "98b1140803dcb99a4a48cf8f14fc099961ad55f7";
+    private final static String version = "a85001e8e93a1271ccfc8edb0810c07041f24661";
 
     /**
      * flag so that initialization in only done once
--- a/de.peeeq.wurstscript/src/test/java/tests/wurstscript/tests/StdLibOwnTests.java
+++ b/de.peeeq.wurstscript/src/test/java/tests/wurstscript/tests/StdLibOwnTests.java
@@ -69,4 +69,36 @@ public class StdLibOwnTests extends WurstScriptTest {
         // count: it is there to catch the program holding none, not to be updated on every bump.
         test().withStdLib().expectAtLeastTests(100).lines(program.toArray(new String[0]));
     }
+
+    /**
+     * A failing library test says which one it was.
+     * <p>
+     * The name is printed to stdout, which a CI run does not keep, so the thrown message was all that
+     * was left of it - and it carried the counts followed by every warning the library compiles with.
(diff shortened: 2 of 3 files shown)


Frotty:
@codex review

Frotty:
Pin bumped to `a85001e` in c248a9035, which carries WurstStdlib2#467 — so this now fixes the macos-15-intel failure rather than only explaining it. 460/460 against the new pin, with the benchmark reporting 1152 lookups instead of 10000. Title and description updated to match.

@codex review

Frotty:
Qualified in e33285482. `RunTests` already composes `package.name` for its own output, so this uses that rather than a second spelling of it, and the test now asserts `test.deliberatelyFails` rather than the bare name. You are right that it mattered: `standardLibraryTestsPass` imports every test package the library has, so two of them naming a test the same would have been indistinguishable.

`StdLibOwnTests` green.

@codex review

Frotty:
Description updated to mention the package qualification from e33285482, which landed after it was written.

@codex review

PR #1255, opened 2026-08-18
Write down the natively keyed store design before building it

Frotty:
Backlog item 27, agreed in discussion. No code — this records a design and the groundwork for it, so the implementation does not start by rediscovering what was already established.

### The decision

A Lua table is already a hash map, so a bounded store should lower to `t[key] = value` there rather than hashing and probing. That would lift `FASTHASHMAP_CAPACITY` / `FASTHASHMAP_MAX_INSTANCES` on that target, and get string keys off `StringHash` — which is case insensitive (`String.wurst:81`), collapses every partial multibyte slice to one constant (`MultibyteDiagnostics`), and is undocumented and changed between game versions.

The interpreter *does* emulate it — `StringProvider` delegates to `Wc3StringHash`, and `Wc3StringHashTest` checks whole strings, partial multibyte bytes and parity with the Lua shim — so a test does see the real behaviour. An earlier revision of this note claimed otherwise and was wrong; what makes `StringHash` unusable is the behaviour itself, not the fidelity of the emulation. (`MultibyteDiagnostics`' own comment is stale on this point in the library too.)

### Why a type class decides it

A Lua table matches keys by raw identity — reference identity for a class. An instance whose `equals` is structural, `Hashable<vec2>` comparing components, would have Jass treat two equal-valued keys as one key and Lua treat them as two, from one program, silently. So the native path is sound only for `int`, `real`, `string`, `boolean` and reference-keyed classes, and a requirement-free bound states which those are:

```wurst
public interface RawKeyed<T:>
```

**How the two variants are selected is unsettled, and the obvious spelling does not compile.** Wurst does not overload a type on its bounds, so declaring `FastHashMap<K: Hashable>` beside `FastHashMap<K: Hashable and RawKeyed>` makes every mention of the name ambiguous before any instance is considered. Either the native variant is a separate type — `RawHashMap`, with the bound as its entry condition — or one type carries both strategies and branches per operation, which costs a branch and gives up the limits being lifted. That is the first thing to decide.

A further constraint: `null` is a value of any class type and lowers to `nil`, and `t[nil]` is a runtime error in Lua while the probing implementation accepts it wherever `Hashable` does. Identity alone is therefore not enough to admit a reference-keyed class.

### Groundwork recorded

- Intrinsics are declared in Wurst with `@compilerintrinsic` and bounds parse there; `wurstNewInstance<T:>() returns T` is the shape. Recognition is name plus `!AttrFuncDef.hasApplicableUserFunction`.
- `ImTranslator.isLuaTarget()` is available during Wurst-to-IM lowering. This is what makes it feasible — an `if isLua` guard alone does not, because folding runs after translation and both branches get lowered.
- The Lua backend already emits a plain `t[i]` for an array access (`translateArrayAccessRaw`). Nothing there needs changing; only the language's `int` index requirement blocks this, and typechecking is target independent.
- `ImTranslator.imError` covers a Jass lowering with no Jass meaning; `ImStatementExpr` pairs statements with a value.

### Left open on purpose

The read is straightforward. The write has to become a statement while an `ExprFunctionCall` must lower to an `ImExpr` — either an `ImStatementExpr` with a discarded value, or an AST expansion after validation the way `wurstMapFields` assigns back to fields. Guessing that is how this would go wrong.

### Status

WurstStdlib2#468 is no longer blocked on this. The correctness motive is gone — `FastHashMap` computes its own hash now, so the `StringHash` defects do not reach it, and it behaves the same on both targets. What is left here is performance and the capacity limits on Lua.

### DIFF
--- a/BACKLOG.md
+++ b/BACKLOG.md
@@ -107,6 +107,52 @@ itself, and one gap in what the suite can see.
     `FASTHASHMAP_CAPACITY`, `isFull()` and `Hashable.hash` to the public API when the Lua path makes
     all three meaningless on that target.
 
+28. **Two costs the Lua emission pays which look avoidable.** Found by reading the emitted script
+    for `FastHashMapTests_fastHashMapRuntimeLua`, not by profiling. **Measure both before touching
+    either**; each proposal below is a structural argument and a call count, which is not the same as
+    knowing what the game's interpreter charges for it.
+
+    *Every primitive array read is a function call.* `LuaNativeLowering.lowerPrimitiveArrayEnsure`
+    wraps each non-lvalue primitive array access in `ensureInt`/`ensureBool`/`ensureReal`, and the
+    wrapper survives the optimiser into the emitted script:
+
+        if not(__wurst_ensureBool(FastHashMap_used[s])) then
+        return ((s2 >= this5.FastHashMap_base) and __wurst_ensureBool(FastHashMap_used[s2]))
+
+    Writes are raw, lvalues being skipped, so only reads pay. `slotFor` reads `used` and `dead` every
+    probe step, which is two Lua calls per step in the hottest loop the container has. The shim itself
+    is needed: an untouched table key is `nil` where Jass reads `0` or `false`, and class and handle
+    typed arrays already skip it because `nil` is the right default for them.
+
+    The fix is to move the default off the access site and onto the table, which is the ordinary Lua
+    idiom for exactly this:
+
+        __wurst_default_false = ({__index = function() return false end})
+        setmetatable(FastHashMap_used, __wurst_default_false)
+
+    A present key is then a raw table index with no call at all, and the function runs only on a miss
+    - which is the rare case, a read of a slot never written. One metatable per primitive array global
+    replaces a wrapper at every read site. Three shared metatables cover int, bool and real. `pairs`
+    is unaffected by `__index`, so nothing which iterates an array changes. What has to be checked:
+    where the `setmetatable` calls are emitted (globals init, before any read), that a `0` or `false`
+    actually stored still reads back rather than falling through, and whether anything relies on the
+    stacktrace argument `callWithStacktrace` adds to the current wrapper.
+
+    *A method with an override stops being a direct call.* Without one, a bounded generic's method
+    lowers to a direct call - `Box_render_specialized(Box_new_Box(), 42)`. Add a subclass which
+    overrides it and the same call becomes `b:Box_size_specialized_integer(1)`, an instance miss
+    followed by an `__index` hop to the class table. The tables are flat rather than chained -
+    `SubBox.Box_size_specialized_integer` is assigned on `SubBox` itself, so it is one hop and not a
+    walk up the hierarchy - so the cost is that hop plus a dynamic lookup against a global function
+    call, and adding an override anywhere silently converts every call site of that slot.
+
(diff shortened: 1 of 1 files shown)


Frotty:
Both corrected in a24036216, and the null point recorded.

**Selection mechanism.** You are right that the sketch does not compile — Wurst does not overload a type on its bounds, so `FastHashMap<K: Hashable>` beside `FastHashMap<K: Hashable and RawKeyed>` makes every mention of the name ambiguous before any instance is considered. The note now records that the selection is unsettled and the two shapes it could take — a separate type with the bound as its entry condition, or one type carrying both strategies and branching per operation, which costs a branch and gives up the limits being lifted — rather than a spelling that would not build.

**`null` keys.** Recorded as a constraint on the native path: `null` is a value of any class type, lowers to `nil`, and `t[nil]` is a runtime error in Lua while the probing implementation accepts it wherever `Hashable` does. So identity is not on its own enough to admit a reference-keyed class.

**Interpreter and StringHash.** Checked and you are right: `StringProvider` delegates to `Wc3StringHash`, and `Wc3StringHashTest` covers whole strings, partial multibyte bytes and parity with the Lua shim. The claim is removed. The case insensitivity and the multibyte collapse are real and remain the reason not to use it — what was wrong was saying a test could not see them. I had taken that from `MultibyteDiagnostics`, whose comment predates the emulation; worth knowing that comment is now stale in the library too.

@codex review

Frotty:
Description brought in line with a24036216 — it still carried both claims you corrected: the two ambiguous `FastHashMap` declarations, and the assertion that the interpreter does not emulate `StringHash`. Both are fixed in the note and now in the description, along with the `null` constraint.

Also updated the status: WurstStdlib2#468 is no longer blocked on this. `FastHashMap` computes its own hash now, so the `StringHash` defects do not reach it and it behaves the same on both targets — what remains here is performance and the capacity limits on Lua, not correctness.

@codex review

Frotty:
Two more things recorded in d26c0dc1a, both from discussion, and both change what this entry leads to.

**Which target the container is for.** The read that `FastHashMap` is really a Lua container is right about speed and not about usefulness. On Jass a native `Table` hashes outside the script, so probing arrays with a hash computed in Wurst will not beat `HashMap` — the name only earns itself on Lua. But `HashMap` cannot take a `vec2`, a tuple, or anything not castable, and two keys which cast to one int collide there; that niche is target independent and is why the container exists. So the direction recorded is: keep both, make the native table the *primary* implementation rather than an optimisation, and stop presenting the Jass path as the fast one. That also settles the selection question this entry had open, because the two variants stop being peers — one becomes the fallback.

**A `Hashing` package**, so instances and future containers stop hand-rolling a mix each. It runs into the same divergence, sharply. `fmix32`, xxHash and FxHash are all xor, shift and wrapping multiply, and `Bitwise.bwXor32` extracts eight bytes and does four table lookups *per xor* — so `fmix32` would cost roughly twenty four divisions and twelve lookups per integer on Jass, worse than the arithmetic mix it would replace, for avalanche nobody observes at `mod 32`. On Lua those ops are native and it is the right answer outright. It also depends on 32-bit wrapping multiply, which Jass has and Lua does not, so identical values across targets need masking — two more `and32`. One surface, `hashInt`/`hashString`/`combine`, split by target on the same `isLuaTarget()` lowering this entry is about, which makes the package a second consumer of that work rather than separate work.

Both notes say to measure `bwXor32` before committing to any of it. Its cost is the whole argument and I read it from the source rather than timing it.

@codex review

PR #1260, opened 2026-08-28
Require explicit build optimization flags

Frotty:
## Summary

- stop Build Map from unconditionally enabling `opt`, `inline`, and `localOptimizations`
- require each build-only optimization to have its corresponding `+` entry in `wurst_run.args`
- update the regression test to prove an omitted `+opt` stays disabled

## Acceptance criteria

- `+inline` enables inline only for builds
- `+localOptimizations` enables local optimizations only for builds
- omitted `+opt` does not enable output optimization

## Checks

- `./gradlew.bat test --tests tests.wurstscript.tests.WurstCommandsTests` — passed
- `git diff --check` — passed

## Known gaps

- none


Frotty:
@codex review

PR #1269, opened 2026-08-29
Keep tuple vararg loop bindings rooted through lowering

Frotty:
## Summary

- represent vararg-loop bindings explicitly in IM instead of retaining one tuple variable reference
- expand tuple-valued loop bindings to their ordered scalar locals during tuple elimination
- teach copying, rooting checks, liveness, side-effect analysis, local merging, inlining, the interpreter, and Lua emission about the explicit bindings
- add a focused optimized Lua regression for a generic vararg method specialized with a tuple of handles

## Root cause

Lua retains `ImVarargLoop` until backend emission. Tuple elimination replaced the tuple local in the function's locals with scalar locals, but the loop still referenced the detached tuple variable. Local optimization then inspected that stale variable and failed with `Variable ... not attached`.

The IM now owns a list of binding nodes which reference the live locals. Before tuple elimination the list has one binding; tuple elimination replaces it with one binding per scalar leaf. Lua consumes that lowered representation directly instead of reconstructing tuple membership from translator state.

## Verification

- `./gradlew.bat compileJava`
- `./gradlew.bat compileTestJava`
- `./gradlew.bat test --tests "tests.wurstscript.tests.LuaBackendAuditTests.optimizedTupleVarargLoopUsesAttachedScalarLocals" --tests "tests.wurstscript.tests.VarargTests"`
- `./gradlew.bat shadowJar`
- Castle Fight PR 65 map build using the resulting `wurstscript.jar`: passed with exit code 0; no tracked Castle Fight files changed

The full suite was intentionally left to CI.

Context: https://github.com/Frotty/wurst-castle-fight/pull/65

### DIFF
--- a/de.peeeq.wurstscript/parserspec/jass_im.parseq
+++ b/de.peeeq.wurstscript/parserspec/jass_im.parseq
@@ -74,6 +74,9 @@ ImMethod(@ignoreForEquality de.peeeq.wurstscript.ast.Element trace,
 	
 	
 ImStmts * ImStmt
+ImVarargLoopVars * ImVarargLoopVar
+
+ImVarargLoopVar(ref ImVar var)
 
 ImStmt = 
 	  ImIf(@ignoreForEquality de.peeeq.wurstscript.ast.Element trace, ImExpr condition, ImStmts thenBlock, ImStmts elseBlock)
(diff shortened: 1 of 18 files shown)


Frotty:
@codex review

PR #1282, opened 2026-09-02
Reduce transient allocations in hot compiler passes

Frotty:
## Summary

- bump `abstractsyntaxgen` from `623da1c60f` to `2b3d742e8a` on both the compiler and AST-generation classpaths
- replace allocation-producing iterator, stream, lambda, and collection-copy traversals with indexed access across frequently repeated IM passes
- avoid materializing combined function collections when flattening, merging locals, analyzing local-player context, and tracing Lua reachability
- preserve the required iterator for the `HashMultimap` removal view, where indexed traversal is not available and taking a snapshot would allocate more

## Affected compiler paths

- IM flattening and helper traversal
- local-variable liveness and merging
- local-player dependency analysis
- Lua dispatch preparation
- Lua garbage collection and generic-static initializer cleanup

## Motivation

These passes repeatedly traverse generated AST/IM list types across the whole program. Using their indexed access directly removes avoidable transient iterators, streams, snapshots, and aggregate sets without changing traversal order or compiler semantics.

## Validation

- Gradle build and test CI passed
- published test results passed
- Codex reviewed head `51456777b3` and reported no major issues

No source-language or backend behavior change is intended.

### DIFF
--- a/de.peeeq.wurstscript/build.gradle
+++ b/de.peeeq.wurstscript/build.gradle
@@ -98,8 +98,8 @@ dependencies {
     implementation "org.antlr:antlr4-runtime:4.13.1"
 
     // abstractsyntaxgen (available to IDE via compileOnly; used at runtime via astgen)
-    compileOnly 'com.github.peterzeller:abstractsyntaxgen:623da1c60f'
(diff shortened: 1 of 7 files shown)


Frotty:
@codex review

PR #1293, opened 2026-09-04
Fix dangling Lua enum helper calls

Frotty:
## Summary

- use WC3's native enum accessors while Lua context collectors are executing inside native callbacks
- avoid dangling references to separately garbage-collected `__wurst_GetEnum*` helpers
- cover force, group, item, and destructable collector output

## Acceptance criteria

- `ForForce` works when user code does not otherwise reference `GetEnumPlayer`
- generated context collectors contain no hidden dependency on separately emitted helper functions
- override-aware helpers remain available for replayed callbacks that call `GetEnum*`

## Checks

- focused Lua callback-helper regression set: passed
- `LuaTranslationTests`: passed
- full `test` suite: passed (completed before the request to leave the full suite to CI)
- `git diff --check`: passed

## Known gaps

- no Warcraft III client run; generated Lua and Lua runtime tests cover this compiler-emission defect


Frotty:
@codex review

PR #1295, opened 2026-09-04
Match fuzz budgets to generator shapes and assert real invariants

Frotty:
The compiler fuzz generator varies program structure with six bits of the seed (indent style, tuple, interface, module, loop, callback), giving 64 distinct shapes for single programs and 24 for cross-package ones; the rest of the seed only renames packages and changes integer literals, which reaches no new compiler path. The invocation budgets were 320/180/120 against those counts, so most invocations recompiled shapes already covered, and three of the four properties asserted only that the result was non-null.

Budgets now derive from named constants tied to what each generator can actually produce, and the weakest property is replaced by two that assert real invariants:

- `newlineStyleDoesNotAffectEmittedCode` - the same source written with LF and CRLF must emit byte-identical Jass (was: compiling the CRLF variant returns non-null).
- `compilingTwiceEmitsIdenticalCode` - recompiling the same source in one process must emit byte-identical Jass, across every generated shape rather than the handful `DeterministicChecks` pins.

Compilations drop from 684 to 408. `SmallCheckViaJUnitCoreTestNG` falls from 360.4s to 42.0s in isolation and 129.8s under full parallel load, and is no longer the suite's slowest class.

Checks: full suite green locally, 1941 tests, 0 failures, 6m34s (was 7m17s on the base branch).

Stacked on #1294 - `compileAndReadJass` uses `testNamed`, which is added there. Retarget to master once that merges.

### DIFF
--- a/de.peeeq.wurstscript/build.gradle
+++ b/de.peeeq.wurstscript/build.gradle
@@ -13,6 +13,7 @@ plugins {
     id 'eclipse'
     id 'idea'
     id 'jacoco'
+    id 'net.ltgt.errorprone' version '5.1.1'
     id 'maven-publish'
     id 'com.gradleup.shadow' version '9.2.2'
     id 'de.undercouch.download' version '5.6.0'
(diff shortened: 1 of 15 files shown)


Frotty:
@codex review

Frotty:
@codex review

